PR #1, opened 2021-01-21
[TextField] add secret prop (bool) for password field

lucatr:
Setting prop secret to true applies type 'password' to the input field, i.e. entered text will be obscured. Defaults to false, which will apply type 'text'.


neonnoon (approved):
Thanks, looks good to me!

OSAGfar (approved):

PR #2, opened 2021-02-09
[RadioGroup] add optional title

lucatr:
Added optional title to radiogroup element, as available for other form elements, such as textfield or textarea. Unlike radiobutton or checkbox, IMO radiogroup is a independent form element which often requires a title.


neonnoon:
Looks good to me! ✌️

neonnoon (approved):

PR #4, opened 2021-02-18
[Button] add prop to set button type attribute

lucatr:
Naming of props, i.e. 'type' vs. 'actionType', is not ideal. However, with the selected naming it is a non breaking change.
Alternatively we could have button type 'submit' implicitly set via style type 'primary'. I decided in favour of flexibility, i.e. allow 'reset' button and uncouple style from function.


neonnoon:
Hmm yeah, good point. In addition to your suggested options, I see a third possibility:
We change/extend the `type` prop to cover certain use cases which imply button type and its style, e.g. `type=submit` = primary + button type submit. This could be cool because it prevents usage of anti-patterns like submit with secondary button. But of course there is a certain risk that (a) this pollutes the type prop and (b) prevents usage of certain valid cases.
This is similar to your suggested alternative, but rather than implying the type, we'd be implying the style. Maybe @ibaldi can help us with what combinations make sense from a UX perspective?

For your use case, we could just create a type=reset which sets html button type to reset and uses secondary styling.

pyrene/src/components/Button/Button.tsx
```diff
@@ -50,7 +56,8 @@ const Button: React.FC<ButtonProps> = ({
 }: ButtonProps) => (
   <div styleName="buttonContainer">
     <button
-      type="submit"
+      // eslint-disable-next-line react/button-has-type
```
neonnoon, line 59: Why is this needed? Does the rule not detect the default value?

neonnoon:
@lucatr, still interested in this? 😄

pyrene/src/components/Button/Button.tsx
lucatr: No, sonar complained about this.

lucatr:
> @lucatr, still interested in this? 😄

Yes, sorry, lost track of this. 
I like your suggestion, but shouldn't we change the default type to button then? Do you think that could break something?

neonnoon:
We would definitely have to keep the "primary"/"secondary" etc types as well, then I don't see how it should break anything?
primary = primary/submit
secondary = secondary/submit
...
reset = secondary/reset
submit = primary/submit

What do you think?

lucatr:
In my case I'm having multiple buttons in a form. Currently I'm unable to control which one is the default submit, for implicit submission [1]. Of course I could work around this by making my submit button the first one in the tree and some CSS. However, to me this feels a bit hacky.
Thinking about it again, I'm not sure if introducing further types, which are possibly redundant (primary = primary/submit and submit = primary/submit) is the way to go. Also I'm concerned that binding style and button type could prevent valid use cases. I'm therefore in favour of a dedicated prop (maybe formAction) which defaults to 'submit'. This wouldn't break anything and also be in line with the button html element default, afaik.

However, I can work around this. If you feel like this pollutes the component, I'm ok with ditching this.

[1] https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

neonnoon:
@ibaldi , what are your thoughts about coupling style type to function? E.g. submit is always primary, cancel is always secondary?

lucatr:
@ibaldi any inputs from you side? Is there a usecase where any other button than primary would be used to submit a form? This also goes a bit into the accessibility topic.

ibaldi:
So here are my 2 cents....

I am not in favor of tying style type to function. Things can change and we would keep flexibility if we don't.

neonnoon (approved):

neonnoon:
Alright, let's go with this then. 😆 
Sorry for the delay, @lucatr .

PR #5, opened 2021-03-14
Do not capitalize text for links.

neonnoon:
This is similar to 0afff44.


neonnoon:
There's quite some other components that capitalize as well, and probably shouldn't:

```
❯ fgrep -r 'text-transform: capitalize' src/components
src/components/LabelAndValue/labelAndValue.css:  text-transform: capitalize;
src/components/Checkbox/checkbox.css:  text-transform: capitalize;
src/components/KeyValueTable/keyValueTable.css:  text-transform: capitalize;
src/components/TabView/tabView.css:        text-transform: capitalize;
src/components/Container/container.css:      text-transform: capitalize;
src/components/TreeTable/TreeTableHeader/treeTableHeader.css:  text-transform: capitalize;
src/components/Table/table.css:  text-transform: capitalize;
src/components/Filter/FilterOption/filterOption.css:  text-transform: capitalize;
src/components/DropdownButton/dropdownButton.css:  text-transform: capitalize;
```

But let's focus on links only to make things easier for now.

neonnoon:
Failing tests is unrelated (broken on main) and should be solved with #6 .

lucatr:
Makes sense. I'm also in favour of removing all text-transform from components.

lucatr (approved):

nicholas-os:
It's the right thing to do, but we will need to prepare for this. All links have to be adapted.

nicholas-os:
Would it make sense to have a `uppercase` property?

neonnoon:
I think we discussed this in the PR for 0afff44 and decided to go for the "breaking change" and not doing any capitalisation. In my opinion, there is no real use case for capitalisation, with different languages and styles (title case vs. sentence case) automatic handling will be wrong more often than not.

neonnoon:
As discussed, we'll leave as is for now and check internal references first.

justmejulian:
@Sakai-san and @sanmiawi is this still needed or can it be declined/merged?

neonnoon:
Superseded by #35

PR #6, opened 2021-03-17
Fix time range selector timezones

neonnoon:
This PR removes all uses of `date-fns-tz` (except from test code). In my opinion, `date-fns-tz` actually complicates things significantly. From a time range point of view, there are no timezones, there's a start and end timestamp. So, the timezone only becomes relevant for display.

`Intl.DateTimeFormat` [1] handles this well. In my opinion, `date-fns-tz` can be useful to handle string to timestamp if a timezone is involved, i.e., figuring out the timestamp of "17 Mar 11:05".

If I understood things right, there's actually three timezones involved, local time, utc, and zoned time. Having the browser (or the tests) run in Europe/Zurich looks correct, using GMT though shows the discrepancy.

[1] https://caniuse.com/?search=DateTimeFormat

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '14'
     - name: Lint pyrene
       run: (cd pyrene && npm install && npm run lint)
   tuktuktwo:
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '14'
     - name: Lint tuktuktwo
       run: (cd tuktuktwo && npm install && npm run lint)
   pyrene-graphs:
@@ -31,7 +31,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '14'
     - name: Lint pyrene-graphs
       run: (cd pyrene-graphs && npm install && npm run lint)
   kitchensink:
@@ -41,6 +41,6 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v1
       with:
-        node-version: '12'
+        node-version: '14'
     - name: Lint kitchensink
       run: (cd kitchensink && npm install && npm run lint)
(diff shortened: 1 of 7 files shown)


neonnoon:
Before, correct when browser is in CET (start of next interval is end of previous):
before-cet-cest before-cet-cet

Before, incorrect when browser time is in GMT (one hour missing in DST crossing):
before-gmt-cest before-gmt-cet

After, issue is fixed when browser time is in GMT (start of next interval is end of previous):
after-gmt-cest after-gmt-cet

For completeness sake, correct behaviour also when browser is in CET:
after-cet-cest after-cet-cet

neonnoon:
Updated node from `12` to `14`. v12 did not yet contain full ICU support by default, see:
https://github.com/nodejs/node/issues/19214

nicholas-os:
Thank you Mezzo!

neonnoon:
Thanks for the review, @nicholas-os . @justmejulian , do you also wanna have a quick look?

neonnoon:
Discussed with @sosostris , @justmejulian & Thomas, merging now.

Sakai-san (commented):
you can do destructuring here : 

`const TimeRangeNavigationBar = (props) => {`

`const TimeRangeNavigationBar = ({from, lowerBound}) => {`

justmejulian (commented):
Looks loads cleaner and fixes the failing tests. Thanks!

sosostris:
@neonnoon I merged all the changes to local repo and somehow the tests were failing due to formatting (the values are all correct), like this:

````
Expected: "01.11.2020, 01:59 - 02.11.2020, 00:59"
Received: "11/01/2020, 01:59 AM - 11/02/2020, 12:59 AM"
````

But the tests are all passing here, so it's super weird. Checked a bit together with @Sakai-san and we found that the implementation of the method Intl.DateTimeFormat seems to vary with machine-os-language-set-up, e.g. there are similar issues reported here https://stackoverflow.com/questions/52797070/intl-datetimeformat-with-format-options-doesnt-work-for-it-ch

I will create a ticket to follow up with this. It would be nice to unify all the display to be one format :)

neonnoon:
Ah.. I had the same issue.

See build result: See https://github.com/open-ch/pyrene/runs/2129628724 
Fixed for Github like so: fa7937484d2861206f0b8b9cc77a1c8aa48f300f
Also mentioned further up in this thread: https://github.com/open-ch/pyrene/pull/6#issuecomment-800977148

You might have to check the bamboo build agents.

PR #8, opened 2021-03-30
Allow react 17 as peer dependency

neonnoon:
This is only to allow react 17 in projects together with pyrene, it does not update which react version is actually used.


neonnoon:
Again.. test will fail due to #6 .

nicholas-os (approved):

lucatr (approved):

PR #12, opened 2021-04-08
WIP/Proposal: ThemeProvider sets CSS custom properties for themeable styles

neonnoon:
This is a proposal rather than the full implementation:

https://user-images.githubusercontent.com/549158/114041215-a0741800-9884-11eb-866d-877e6c0c249f.mov

Basic idea:

1. CSS custom properties (a.k.a. variables) can be used natively in supported browsers
2. `ThemeProvider` provides those variables for all components to use (most components already use them)
3. For the time being, we should keep `colors.css` which is basically "the default values", which also means that pyrene keeps working as is without theme provider

Next steps:
1. Both default theme and dark theme should be provided as exported constants
2. Investigate color constants / colors.css and how duplication can be reduced once we have ThemeProvider

What do you think?

### DIFF
--- a/kitchensink/.babelrc
+++ b/kitchensink/.babelrc
@@ -1,10 +1,6 @@
 {
     "presets": [
-      ["@babel/preset-env", {
-        "targets": {
-          "browsers": [ ">0.1%", "ie 11", "not op_mini all"]
-        }
-      }],
+      "@babel/preset-env",
       "@babel/preset-react"
     ],
     "plugins": [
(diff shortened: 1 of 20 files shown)


neonnoon:
Could you have a look at this and let me know what you think of the general approach?
If you think it's not complete nonsense, I'd continue with the implementation.

neonnoon:
Closing this for now, will discuss future of this offline and reopen if appropriate.

PR #13, opened 2021-04-13
Improvement - remove displayName debugging property - WIP

Sakai-san:
## Done

1. **displayName**
`diplayName` is already included when using `FunctionComponent` type generic. See [React+TypeScript Cheatsheets](https://github.com/typescript-cheatsheets/react#function-components) below.

___
Screenshot 2021-04-13 at 08 33 39
___
2. **Type inference**
Let TypeScript do the **type inference** whenever possible.
3. **Eslint**
`react/prop-types` eslint rule **disabled** for React component in TypeScript.


pyrene/.eslintrc
```diff
@@ -15,5 +15,13 @@
     "react/forbid-prop-types": 0, 
     "no-multiple-empty-lines": 0,
     "@typescript-eslint/unbound-method": 0
-  } 
+  },
+  "overrides": [
```
neonnoon, line 19: Shouldn't this better be part of the `eslint-config` package?
No strong opinion though.

pyrene/src/components/Accordion/Accordion.tsx
```diff
@@ -30,6 +30,5 @@ const Accordion: React.FC<AccordionProps> = ({
   </div>
 );
 
-Accordion.displayName = 'Accordion';
```
neonnoon, line 33: Does this work with kitchensink? I think we used to read `displayName` for the component name there. If it's compatible, all good. If not, we might need minor changes there.

Generally though, we should rather tweak kitchensink to support a cool developer experience in pyrene than the other way round (i.e., require anti-patterns in pyrene just because of kitchensink).

neonnoon (changes requested):
Selecting *Request changes*, but if kitchensink works and you prefer the eslint changes in .eslint rather than config package. All good for me. ✌️

pyrene/src/components/Accordion/Accordion.tsx
neonnoon: I'm a bit confused which PR is which, and can't find your comment with the `compareLocale` issue.
However, I am pretty sure that it is related to this comment. Kitchensink currently takes the component name from `displayName` if you remove that, the name has to be taken from somewhere else.

PR #15, opened 2021-04-13
Improvement/no-js-interpretation-in-jsx

Sakai-san:


neonnoon (approved):
How did this even happen? 😃 

@neonnoon The current code still works. Sometimes while you are coding you start with `{ items.map() ...}` in JSX and you realise, ooh I can actually reuse that component. And you forget to remove the `{}`.

PR #17, opened 2021-04-14
Checkbox list in typescript

Sakai-san:


neonnoon:
@justmejulian , I think we discussed this at some point, didn't we? The way TypeScript has been introduced was not really thought through, but rather a bit forced. We started bending pyrene implementation to still be compatible with kitchensink, whereas we should have rethought how we extract prop information from pyrene. I don't think there's a quick and dirty fix, we'll have to reassess possibilities.

There might be alternatives to https://github.com/storybookjs/babel-plugin-react-docgen, e.g., https://github.com/strothj/babel-plugin-react-docgen-typescript (although looks half abandoned and little used)? But given the popularity of babel-plugin-react-docgen and TypeScript, I'm sure we're not the only one with the issue.

By the way, please also note that `CheckboxPopover` is a bad example to compare before and after, because it doesn't make use of comments. See e.g., [Checkbox](https://open-ch.github.io/pyrene/Form/Checkbox) where prop and component description is extracted from comments in the dev build.

Sakai-san:
> @justmejulian , I think we discussed this at some point, didn't we? The way TypeScript has been introduced was not really thought through, but rather a bit forced. We started bending pyrene implementation to still be compatible with kitchensink, whereas we should have rethought how we extract prop information from pyrene. I don't think there's a quick and dirty fix, we'll have to reassess possibilities.
> 
> There might be alternatives to https://github.com/storybookjs/babel-plugin-react-docgen, e.g., https://github.com/strothj/babel-plugin-react-docgen-typescript (although looks half abandoned and little used)? But given the popularity of babel-plugin-react-docgen and TypeScript, I'm sure we're not the only one with the issue.
> 
> By the way, please also note that `CheckboxPopover` is a bad example to compare before and after, because it doesn't make use of comments. See e.g., [Checkbox](https://open-ch.github.io/pyrene/Form/Checkbox) where prop and component description is extracted from comments in the dev build.


@neonnoon Thanks for sharing your thoughts. The first babel plugin you mentioned can be used if you intend to keep the `propTypes` attribute in React components. But migrating to TypeScript means **removing** it. Otherwise you would have **both** static and run time type checking. Static type checking is better since it prevent issue upstream in contrast to discover bugs while actually using the application. We ll let you know how we d like to tackle this issue.

neonnoon:
Sorry, I also kinda lost context here. Do I understand things right that 1a166608d95a9e751c8dd31d79e664b4ae00a2bc was all that's needed for kitchensink to understand prop interfaces?

Sakai-san:
> Sorry, I also kinda lost context here. Do I understand things right that [1a16660](https://github.com/open-ch/pyrene/commit/1a166608d95a9e751c8dd31d79e664b4ae00a2bc) was all that's needed for kitchensink to understand prop interfaces?

@neonnoon You are right ! Also you I remove the **type combination** (algebraic data type) which does not work with `Kitchensink`.

neonnoon (approved):

PR #26, opened 2021-04-20
Container component in TypeScript

Sakai-san:


pyrene/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,154 @@
...
+    default: {
+      return { ...state };
+    }
+  }
+};
+
+const Container: FunctionComponent<ContainerProps> = ({
+  collapsible = false,
+  defaultExpanded = false,
+  onChange = () => null,
```
justmejulian, line 91: can't this just be undefined?

pyrene/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+interface LoadingAction extends Action {
+  type: 'loading',
+  payload: {
+    contentHeight: number | null
+  }
+}
+
+const reducer = (state: State, action: LoadingAction | TogglingAction | ChangingAction) => {
```
justmejulian, line 66: 😎

pyrene/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,154 @@
...
+        },
+      });
+    }
+  };
+
+  return (
+    <div className={clsx(styles.container, { [styles.expanded]: state.expanded || !collapsible })}>
+      <div className={clsx(styles.header, { [styles.collapsible]: collapsible })} onClick={toggleCollapse} role="button" aria-label="Show or hide container">
+        <span className={clsx(styles.title, 'unSelectable')}>
+          {' '}
```
justmejulian, line 133: ❓

pyrene/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,154 @@
...
+  useEffect(() => {
+    if (contentRef?.current) {
+      dispatch({
+        type: 'loading',
+        payload: {
+          contentHeight: contentRef.current.clientHeight,
+        },
+      });
+    }
+  }, []);
```
justmejulian, line 108: does this not need an eslint exception?🤔

pyrene/src/components/Container/Container.tsx
justmejulian: will avoid unnecessary function calls

pyrene/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,154 @@
...
+   */
+  renderCallback: () => JSX.Element,
+  /**
+   * Sets the title displayed in the header of the component.
+   */
+  title: string,
+}
+
+interface State {
+  contentHeight: null | number,
```
justmejulian, line 36: maybe worth moving to a type, because also used [here](https://github.com/open-ch/pyrene/pull/26/files#diff-f3bd19d7822f64efe9ab3b3341869cbc25fadb3c4e45c04c34cf349a7074f2c7R62)

justmejulian: or we make it optional

pyrene/src/components/Container/Container.tsx
Sakai-san: > does this not need an eslint exception?🤔

@justmejulian  This does not raise any problem because we did not integrate the [eslint-plugin-react-hooks](https://reactjs.org/docs/hooks-rules.html#eslint-plugin) in the build process. We should do so, but that is another discussion.

pyrene/src/components/Container/Container.tsx
Sakai-san: > ❓

@justmejulian  This is something I took over from the previous implementation. We can fix that I a separate PR, but I think this aimed to be so.

justmejulian (approved):

PR #31, opened 2021-04-21
Remove text-transform from Container component

sanmiawi:


neonnoon:
I don't quite understand, why is this work in progress? And why do you merge this towards `feature/MC-21315` rather than `main`?

sanmiawi:
> I don't quite understand, why is this work in progress? And why do you merge this towards `feature/MC-21315` rather than `main`?

Well, the idea is that we should have a PR for each change in Pyrene and java-portal. So I felt that it would be easier to organise/follow if it was all done on a feature branch

sosostris (approved):

Sakai-san (approved):

PR #32, opened 2021-04-21
attempt getting rid date-fns-tz in prod build

Sakai-san:
Using @neonnoon [approach](https://github.com/open-ch/pyrene/pull/6/files) for converting a date with `time zone A` to date with `time zone B`, using **built-in JavaScript utility** was a better approach then import[ third-party lib](https://github.com/open-ch/pyrene/pull/19).

### DIFF
--- a/pyrene/package-lock.json
+++ b/pyrene/package-lock.json
@@ -5863,7 +5863,8 @@
     "date-fns-tz": {
       "version": "1.1.4",
       "resolved": "https://registry.npmjs.org/date-fns-tz/-/date-fns-tz-1.1.4.tgz",
-      "integrity": "sha512-lQ+FF7xUxxRuRqIY7H/lagnT3PhhSnnvtGHzjE5WZKwRyLU7glJfLys05SZ7zHlEr6RXWiqkmgWq4nCkcElR+g=="
+      "integrity": "sha512-lQ+FF7xUxxRuRqIY7H/lagnT3PhhSnnvtGHzjE5WZKwRyLU7glJfLys05SZ7zHlEr6RXWiqkmgWq4nCkcElR+g==",
+      "dev": true
     },
     "debug": {
       "version": "4.3.1",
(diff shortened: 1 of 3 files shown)


neonnoon:
Can you elaborate why you prefer going for this variant? Basically you're saying you're trusting browsers to implement `en-us` properly (so you can use it to parse and build the manual date), but not `de`? 

If the only reason is not having node 14 available on the (to be deprecated) internal CI, I don't quite see the point.

Anyway, I think I am missing some context of internal discussions you're having, from my point of view we're going a bit in circles.

Sakai-san:
> Anyway, I think I am missing some context of internal discussions you're having, from my point of view we're going a bit in circles.

Hi @neonnoon. Ok let me give some context. It relates to the last merged PR actually.
The basic idea is to not used third party library while built-in JS object do the same job (like you did in the past).

The language `de`, `en-US` does not matter much. Indeed we do have our own format anyway, namely `'dd.MM.yyyy, HH:mm'`.

The only problem with `de` is that is seems not to be in `node 12`. Important is that the built-in JS feature convert the date into new time zone (including summer time).

The present PR is nothing but a proposal. If there no need I ll close it. All good.

neonnoon:
> The basic idea is to not use third party library while built-in JS object do the same job (like you did in the past).

Exactly, but then we reverted that. And now we're reverting the revert. Which is why I think we're going in circles. 😄

> The language `de`, `en-US` does not matter much. Indeed we do have our own format anyway, namely `'dd.MM.yyyy, HH:mm'`.

`'dd.MM.yyyy, HH:mm'` is actually `de`, just without seconds.

Again, no strong opinion either way. But I think we should come to a conclusion and an end with this topic eventually.

Sakai-san:
> Again, no strong opinion either way. But I think we should come to a conclusion and an end with this topic eventually.

Ok. Cool. Thanks for your feedback. Alles i.O.

neonnoon:
✌️

PR #41, opened 2021-04-26
extended TabView to display extra section alongside tab name

sosostris:
This PR is to extend the existing TabView component such that on the right side of tab name, some additional information could be displayed.

Visual effect:
Screenshot 2021-04-26 at 17 11 52
Screenshot 2021-04-26 at 17 11 56

I originally thought of an alternative approach, which is to make the current `name` prop and `nameRenderCallback` props mutually exclusive - if there is `name`, display the name; otherwise display the customizedly rendered name.

However, after many considerations I switched to the current approach where `name` is displayed no matter what and the additional section on the right side `nameRenderCallback` is optionally rendered.

The reasons are:
- it's backward compatible
- `name` is kinda used as `key` for this component, without the `name` prop we still would need to have something like `key` in its place
- it helps make the CSS more consistent when toggling the tabs; otherwise when using this component, the customized name would have to also care about its different look when it is in selected/unselected state

### DIFF
--- a/pyrene/src/components/TabView/TabView.examples.jsx
+++ b/pyrene/src/components/TabView/TabView.examples.jsx
@@ -1,18 +1,27 @@
 import React from 'react';
+import Icon from '../Icon/Icon';
 import Placeholder from '../../examples/Placeholder';
 
 const TabView = {};
+
+const nameRenderCallback = (name) => (
+  <div style={{ marginLeft: 4 }}>
+    <Icon name={name} />
+  </div>
+);
 
 TabView.props = {
   initialTabName: 'Tab 1',
   directAccessTabs: 3,
   tabChanged: (stateProvider) => () => stateProvider.setState((prevState) => ({ tabName: prevState.tabName ? prevState.tabName + 1 : 1 })),
   tabs: (stateProvider) => [
-    { name: 'Tab 1', renderCallback: () => <Placeholder label="tab 1" />, disabled: false }, // eslint-disable-line react/display-name
+    { name: 'Tab 1', nameRenderCallback: () => nameRenderCallback('home'), renderCallback: () => <Placeholder label="tab 1" /> }, // eslint-disable-line react/display-name
     { name: 'Tab 2', renderCallback: () => <Placeholder label={`Tab ${stateProvider.state.tabName}`} />, disabled: false }, // eslint-disable-line react/display-name
     { name: 'Tab 3', renderCallback: () => <Placeholder label="tab 3" />, disabled: true }, // eslint-disable-line react/display-name
     { name: 'Looooooooooooooooooooooooooooooooooooooong Name', renderCallback: () => <Placeholder label="tab 4" /> }, // eslint-disable-line react/display-name
-    { name: 'Tab 5', renderCallback: () => <Placeholder label="tab 5" />, disabled: true }, // eslint-disable-line react/display-name
+    {
+      name: 'Tab 5', nameRenderCallback: () => nameRenderCallback('info'), renderCallback: () => <Placeholder label="tab 5" />, disabled: true, // eslint-disable-line react/display-name
+    },
   ],
 };
 
--- a/pyrene/src/components/TabView/TabView.examples.jsx
+++ b/pyrene/src/components/TabView/TabView.examples.jsx
@@ -4,7 +4,7 @@ import Placeholder from '../../examples/Placeholder';
 
 const TabView = {};
 
-const nameRenderCallback = (name) => (
+const renderAuxiliaryIcon = (name) => (
   <div style={{ marginLeft: 4 }}>
     <Icon name={name} />
   </div>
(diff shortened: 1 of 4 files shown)


pyrene/src/components/TabView/TabView.jsx
```diff
@@ -112,6 +113,7 @@ export default class TabView extends React.Component {
                 role="tab"
               >
                 {tab.name}
+                {tab.nameRenderCallback && tab.nameRenderCallback()}
```
Sakai-san, line 116: you can also do [that way](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) :

`{tab.nameRenderCallback?.()}`

pyrene/src/components/TabView/TabView.examples.jsx
```diff
@@ -1,18 +1,27 @@
 import React from 'react';
+import Icon from '../Icon/Icon';
 import Placeholder from '../../examples/Placeholder';
 
 const TabView = {};
 
+const nameRenderCallback = (name) => (
+  <div style={{ marginLeft: 4 }}>
```
Sakai-san, line 8: If you d like to, you can horizontally align the tab label and the icon along.

`<div style={{ marginLeft: 4, position: 'relative', top: -2 }}>`


Pyrene

Sakai-san (approved):
This is a great feature. I love icon, so thanks a lot !!

pyrene/src/components/TabView/TabView.jsx
```diff
@@ -189,11 +191,13 @@ TabView.propTypes = {
...
   tabHeaderElement: PropTypes.element,
   /**
-   * Data input array for the tabs. Type: [{ name: string (required), renderCallback: func (required), disabled: bool }]
+   * Data input array for the tabs.
+   * Type: [{ name: string (required), nameRenderCallback: func, renderCallback: func (required), disabled: bool }]
    */
   tabs: PropTypes.arrayOf(PropTypes.shape({
     disabled: PropTypes.bool,
     name: PropTypes.string.isRequired,
+    nameRenderCallback: PropTypes.func,
```
Sakai-san, line 200: You could name that props, but it just from my perspective.

`iconRenderCallback: PropTypes.func,`

Sakai-san (approved):

pyrene/src/components/TabView/TabView.jsx
```diff
@@ -189,11 +191,13 @@ TabView.propTypes = {
    */
   tabHeaderElement: PropTypes.element,
   /**
-   * Data input array for the tabs. Type: [{ name: string (required), renderCallback: func (required), disabled: bool }]
+   * Data input array for the tabs.
+   * Type: [{ name: string (required), nameRenderCallback: func, renderCallback: func (required), disabled: bool }]
```
neonnoon, line 195: `renderCallback: func (required)`, `required` is not correct, is it?

pyrene/src/components/TabView/TabView.jsx
neonnoon: @sosostris , I like your thoughts of always using `name`, it helps with consistency and prevents developers from misusing tabs for weird stuff.

Since the `name` is always present, `nameRenderCallback` can be somewhat confusing (it's not about the name, after all). `iconRenderCallback` is also not quite correct (because no one forces us to use an icon, could be anything).

`descriptionRenderCallback`?, also not quite correct. Of all alternatives I can think of, `nameRenderCallback` still makes most sense. 😝

neonnoon (approved):

pyrene/src/components/TabView/TabView.jsx
sosostris: D'awwww! Indeed! Thanks a lot :) Yes it's more elegant that way :)

pyrene/src/components/TabView/TabView.jsx
sosostris: ood question! let me check :)
Ah, looks like it is indeed required:
`renderCallback: PropTypes.func.isRequired,` :)

pyrene/src/components/TabView/TabView.jsx
sosostris: Hello @Sakai-san @neonnoon 😄  yes !!! I had great difficulty finding a proper name for it 😅  was also a bit concerned it might lead to some misunderstanding ... :( and once we merge the PR it would not be nice to come back to change again ...

The purpose of it is to display some additional quick information relevant to the content of the current tab, so `descriptionRenderCallback` actually better reflects this purpose. I also thought of `renderNameRightSection`,  `renderAuxiliaryInfo`, `additioanlInfoRenderCallback`, 😅  does it ring some bell?

pyrene/src/components/TabView/TabView.jsx
neonnoon: Sorry, my mistake. Confused renderCallback and nameRenderCallback.

pyrene/src/components/TabView/TabView.jsx
Sakai-san: > renderAuxiliaryInfo

I like that one !

PR #42, opened 2021-04-29
Don't use npm ci for kitchensink.

neonnoon:
Reverts part of `npm ci` introduced by e9d2048bf1d7844323c8fc25b232dbca502f52bf. The reason to introduce `npm ci` was exactly to make sure the build fails in case `package.json`/`package-lock.json` changes.

What I have completely forgot about was that we change kitchensink's `package.json` on purpose on pyrene/pyrene-graphs releases. So `npm ci` is not appropriate here.

https://docs.npmjs.com/cli/v7/commands/npm-ci

### DIFF
--- a/kitchensink/package.json
+++ b/kitchensink/package.json
@@ -9,7 +9,7 @@
   "private": true,
   "scripts": {
     "clean": "rm -rf dist node_modules",
-    "prebuild": "npm ci",
+    "prebuild": "npm install",
     "build": "webpack",
     "watch": "webpack-dev-server",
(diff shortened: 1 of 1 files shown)


Sakai-san (approved):
Thanks a lot @neonnoon.

sosostris (approved):
ahh!! i see !! that makes sense !! Thanks a lot :))) 👍

neonnoon:
Looking up unpublish functionality, it seems:
- unpublished versions can never be re-published
- entirely unpublished packages are blocked for 24h.

https://docs.npmjs.com/cli/v7/commands/npm-unpublish

PR #45, opened 2021-04-30
Make pyrene logo work outside fo pyrene.

neonnoon:
@sosostris , FYI.


neonnoon:
Follow up in #46 , logo should show on npmjs after next release.

PR #60, opened 2021-04-30
DropdownButton-in-TypeScript

Sakai-san:


pyrene/src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -1,96 +1,94 @@
...
+  label: string,
+  /**
+   * Disables the component and displays a loader inside of it.
+   */
+  loading?: boolean,
+}
 
-  const onClose = () => setState({ displayActions: false });
+const DropdownButton: FunctionComponent<DropdownButtonProps> = ({
+  actions = [{ label: 'tbd', onClick: () => null }],
```
neonnoon, line 44: Seems like this was also in the JS implementation, but looks weird, doesn't it?

Sakai-san: > Seems like this was also in the JS implementation, but looks weird, doesn't it?

@neonnoon You are right. What we can do is to create a separate PR for enhancing that. This PR is just about translating from JS to TS.

pyrene/src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -1,96 +1,94 @@
...
-              { loading: props.loading },
-              { openedDropdown: state.displayActions })
+              { disabled: disabled },
+              { loading: loading },
+              { openedDropdown: displayActions })
           }
-          disabled={props.disabled || props.loading}
-          onClick={() => setState({ displayActions: !state.displayActions })}
+          disabled={disabled || loading}
+          onClick={() => setDisplayActions((prevDisplayActions) => !prevDisplayActions)}
```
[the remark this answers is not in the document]
Sakai-san: > Wouldn't this also work?
> `onClick={() => setDisplayActions(!prevDisplayActions)}`

If the update depends on the current, it s [recommended](https://reactjs.org/docs/hooks-reference.html#functional-updates) to use a callback (since it s asynchronous).

PR #73, opened 2021-04-30
change styleName to className in Button component

sanmiawi:


pyrene/src/components/Button/Button.tsx
```diff
@@ -48,25 +48,24 @@ const Button: React.FC<ButtonProps> = ({
...
       type="submit"
-      className="unSelectable"
-      styleName={
-        clsx('button',
-          { [`type-${type}`]: true },
-          { disabled: disabled },
-          { loading: loading })
+      className={
+        clsx('unSelectable', styles.button,
+          { [styles[`type-${type}`]]: true },
```
Sakai-san, line 56: Please do the following when the `class` will be added anyway, means always `true`.

```
styles[`type-${type}`],
```

Sakai-san (changes requested):

PR #82, opened 2021-04-30
change styleName to className in Link component

sanmiawi:


pyrene/src/components/Link/Link.jsx
```diff
@@ -15,7 +15,7 @@ import './link.css';
  */
 const Link = (props) => (
   <a
-    styleName={clsx('link', { [`type-${props.type}`]: true }, { disabled: props.disabled })}
+    className={clsx(styles.link, { [styles[`type-${props.type}`]]: true }, { [styles.disabled]: props.disabled })}
```
Sakai-san, line 18: Please do the following when the `class` will be added anyway, means always `true`.

```
className={clsx(styles.link, styles[`type-${props.type}`], { [styles.disabled]: props.disabled })}
```

Sakai-san (changes requested):

Sakai-san (approved):

justmejulian (approved):

PR #85, opened 2021-04-30
change styleName to className in IconButton component

sanmiawi:


pyrene/src/components/IconButton/IconButton.tsx
```diff
@@ -35,7 +35,7 @@ const IconButton: React.FC<IconButtonProps> = ({
   path = '#',
   type = 'neutral',
 }: IconButtonProps) => (
-  <a styleName={clsx('iconButton', { disabled: disabled }, { [`type-${type}`]: true })}
+  <a className={clsx(styles.iconButton, { [styles.disabled]: disabled }, { [styles[`type-${type}`]]: true })}
```
Sakai-san, line 38: Please do the following when the `class` will be added anyway, means always `true`.

```
<a className={clsx(styles.iconButton, styles[`type-${type}`], { [styles.disabled]: disabled } )}
```

Sakai-san (changes requested):

justmejulian (approved):

PR #86, opened 2021-04-30
change styleName to className in Loader component

sanmiawi:


pyrene/src/components/Loader/Loader.tsx
```diff
@@ -37,13 +37,13 @@ const Loader: React.FC<LoaderProps> = ({
   <>
     {type === 'standalone'
       && (
-        <div styleName={clsx('canvas', { [`size-${size}`]: true })}>
-          <div styleName={clsx('standaloneLoader', { [`styling-${styling}`]: true }, { [`size-${size}`]: true })} />
+        <div className={clsx(styles.canvas, { [styles[`size-${size}`]]: true })}>
```
Sakai-san, line 40: Please do the following when the `class` will be added anyway, means always `true`.

```
<div className={clsx(styles.canvas, styles[`size-${size}`])}>
```

Sakai-san (changes requested):

Sakai-san (approved):

PR #95, opened 2021-05-01
Spacer in TypeScript

Sakai-san:


Leyart (approved):
All good. We should keep in mind to remove the defaultProps from the components once we will migrate to Storybook

PR #102, opened 2021-05-05
Added toggleColumnsCallback to TreeTable

sosostris:
Thanks a lot for having a look at this PR!

The purpose of this PR is to expose a callback function when the columns are toggled for visibility such that the `columns` state could be used by the parent page.

Example use case: when there are multiple tables in different tabs (with TabView), switching to a different tab and then switching back to the original tab would reset all the selected columns in the table, so that user has to select the columns again, which can be quite much effort.

If you find this approach acceptable, I'm wondering maybe it's best that I extend the Table component similarly, so that the behaviors between tables are consistent?

Please let me know what you think :) Thanks a lot!

### DIFF
--- a/pyrene/src/components/TreeTable/TreeTable.jsx
+++ b/pyrene/src/components/TreeTable/TreeTable.jsx
@@ -176,7 +176,7 @@ class TreeTable extends React.Component {
 
     const isColumnHidden = (hidden) => typeof hidden === 'undefined' || hidden !== true;
 
-    const toggleColumnDisplay = (columnId, hiddenValue, callback) => {
+    const toggleColumnDisplay = (columnId, hiddenValue, handler) => {
       const updatedColumns = columns.map((col) => {
         if (col.id === columnId) {
           return { ...col, hidden: hiddenValue };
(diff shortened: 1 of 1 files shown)


pyrene/src/components/TreeTable/TreeTable.jsx
```diff
@@ -413,6 +414,10 @@ TreeTable.propTypes = {
    * Whether the columns (hide/show) popover is available to the user.
    */
   toggleColumns: PropTypes.bool,
+  /**
+   * Callback function when the columns of the table are getting toggled.
+   */
+  toggleColumnsCallback: PropTypes.func,
```
Sakai-san, line 420: @sosostris I get your idea. Look fine to me. I would have called the `props` like the following but it does not matter much.

`toggleColumnsHandler`

Sakai-san (approved):

pyrene/src/components/TreeTable/TreeTable.jsx
neonnoon: Could also be events/callback style with `onColumnsToggle`, but no strong opinion for any of the options. Maybe actually, the `onColumnsToggle` implies too much the `onChange`/controlled pattern?

PR #109, opened 2021-05-14
DropdownButton default props actions

Sakai-san:
Follow-up to that [PR-60](https://github.com/open-ch/pyrene/pull/60/files).

### DIFF
--- a/pyrene/src/components/DropdownButton/DropdownButton.tsx
+++ b/pyrene/src/components/DropdownButton/DropdownButton.tsx
@@ -41,7 +41,7 @@ export interface DropdownButtonProps {
 }
 
 const DropdownButton: FunctionComponent<DropdownButtonProps> = ({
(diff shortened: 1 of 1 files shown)


pyrene/src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -55,7 +55,7 @@ const DropdownButton: FunctionComponent<DropdownButtonProps> = ({
     <Popover
       displayPopover={displayActions}
       onClickOutside={onClose}
-      renderPopoverContent={() => <OptionsList actions={actions} onClick={onClose} />}
+      renderPopoverContent={() => <OptionsList actions={actions || [{ label: 'tbd', onClick: () => null }]} onClick={onClose} />}
```
neonnoon, line 58: Why do we need it here as well? If you don't pass actions, shouldn't there either be an empty dropdown, or an error?

PR #112, opened 2021-05-19
Feature Branch: Components should not capitalize text

justmejulian:
Rebased feature/MC-21315 to not be branched off feature/MC-20600.
Please double check that everything is still here.


sanmiawi:
Looks perfect

PR #115, opened 2021-05-20
Filter in Typescript

Sakai-san:


pyrene/src/components/Filter/FilterPopover/FilterPopover.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onClose = () => null,
+}: FilterPopoverProps) => (
+  <div className={styles.filterPopover}>
+    <div className={styles.title}>Select Filter</div>
+    {negatable && <div className={styles.negateTitle}>Negate</div>}
+    <div className={styles.filterOptions}>
+      {filters.length <= 6
+        ? filters.map((filter) => (
+          <FilterOption {...filter}
+            value={filterValues ? filterValues[filter.id] : null}
```
justmejulian, line 38: Sure null and not undefined?
Null does not allow for/overwrite a default value.

justmejulian (changes requested):

pyrene/src/components/Filter/FilterPopover/FilterPopover.tsx
Sakai-san: > Sure null and not undefined?
> Null does not allow for/overwrite a default value.

@justmejulian Thanks for taking some time for that. In this PR I dit not touch on the logic. So it s just about TypeScript conversion. See the previous file.

Screenshot 2021-05-21 at 09 23 34

justmejulian: Sure but if we find bugs then why not fix them 😊
Not saying it is a bug just saying it might be worth looking at if we are already touching this component.

Also this is probably the reason why [ value?: InputValue | null,](https://github.com/open-ch/pyrene/pull/115/files#diff-e1681eef36baa2d2f1c8ec8af4c7aa3ba704be3768375af68053753ff1e87423L24) was set like that. Why also null and not just optional?
My guess is to allow exactly this usage.

The thing is that proptypes did not complain if you passed null to an optional prop, however typescript does because we have `strictNullChecks` enabled. So why do we hack null as an option to avoid errors that are shown because we use `strictNullChecks`?

Sakai-san: > The thing is that proptypes did not complain if you passed null to an optional prop, however typescript does because we have `strictNullChecks` enabled. So why do we hack null as an option to avoid errors that are shown because we use `strictNullChecks`?

@justmejulian Make sense. Thanks for your valuable remark. Let me check.

justmejulian (approved):

PR #120, opened 2021-05-27
Fix Modal component

sanmiawi:


justmejulian (approved):
Kinda cool, but very hacky solution.
Like discussed, lets keep this as it fixes the issue, but create a follow up issue. incase we choose to not switch to storybook

Sakai-san:
> Kinda cool, but very hacky solution.
> Like discussed, lets keep this as it fixes the issue, but create a follow up issue. incase we choose to not switch to storybook

@sanmiawi @justmejulian Hello guys. Isn't there no more elegant way to solve the problem ? Maybe we can have a little chat about it. I am sure we can come up with something a bit more stable.

sanmiawi:
> > Kinda cool, but very hacky solution.
> > Like discussed, lets keep this as it fixes the issue, but create a follow up issue. incase we choose to not switch to storybook
> 
> @sanmiawi @justmejulian Hello guys. Isn't there no more elegant way to solve the problem ? Maybe we can have a little chat about it. I am sure we can come up with something a bit more stable.

I am open for the discussion. The problem is, it seems to have an "elegant" solution we need to make quite some changes in kitchensink, and the question is, is it necessary to make those changes in kitchensink if we are going to scrap kitchensink?

pyrene/src/components/Modal/Modal.examples.tsx
```diff
@@ -1,6 +1,22 @@
 import React from 'react';
 import Placeholder from '../../examples/Placeholder';
 
+
+const findTriggerButton = () => {
+  const buttons = document.getElementsByTagName('button');
+  const searchText = 'Trigger Modal';
+  const buttonArray = Array.from(buttons);
+
+  return buttonArray.filter((button) => button.textContent === searchText)[0];
```
Sakai-san, line 10: The find function is exactly designed for you :
```
  return buttonArray.find((button) => button.textContent === searchText);
```

Sakai-san (changes requested):

pyrene/src/components/Modal/Modal.examples.tsx
```diff
@@ -1,6 +1,22 @@
 import React from 'react';
 import Placeholder from '../../examples/Placeholder';
 
+
+const findTriggerButton = () => {
+  const buttons = document.getElementsByTagName('button');
+  const searchText = 'Trigger Modal';
+  const buttonArray = Array.from(buttons);
```
Sakai-san, line 8: usually you do plural a variable as convention for developers to know it s about a collection. That way :

```
const buttons = Array.from(buttons);
```

Sakai-san (changes requested):

Sakai-san (approved):

pyrene/src/components/Modal/Modal.examples.tsx
sanmiawi: I would suggest something like 
`const buttonsArray` 
because we already have a 
`const buttons` 
unless we do 
`const buttons = Array.from(document.getElementsByTagName('button'));`

Sakai-san:
> Kinda cool, but very hacky solution.
> Like discussed, lets keep this as it fixes the issue, but create a follow up issue. incase we choose to not switch to storybook

@justmejulian. We ve found a better with @sanmiawi  way to handle this feature.
I'd suggest we delete this PR and merge [that one](https://github.com/open-ch/pyrene/pull/125).

PR #124, opened 2021-05-28
Modal Component Issues - WIP

Sakai-san:
Ticket MC-21501 in terms of OSAG.


sosostris:
@Sakai-san Thanks a lot for the PR :) but instead of saying it's for internal OSAG ticket MC-blah, could you please describe in detail what the issue is about and maybe explain what is the approach to fix the issue :) one thing is it's not good to mention internal OSAG tracking number in a public place, another thing is it gives information to the reviewers who's not OSAG employees ;) 

Also I see in this PR there's a lot of lines that commented out, why is that? :O if the lines are not needed, why not just remove them? 

Thanks a lot!

PR #125, opened 2021-05-29
Modal component can be closed in Kitchensink

Sakai-san:
## Goal of the present PR
In kitchensink while testing the component called [Modal](https://open-ch.github.io/pyrene/Layout/Modal), there is no way to **close** it after that component has rendered.

## Follow-up
The present PR is an alternative to the [PR-120](https://github.com/open-ch/pyrene/pull/120).

## Remark
The present PR implements a React pattern called [Render props](https://reactjs.org/docs/render-props.html).

## Test
Go to `Modal` component in Kitchensink, do trigger the component rendering and hit the key **escape** or click the **close button**.

## Illustration

Screen Recording 2021-05-31 at 07 13 55


sanmiawi (approved):

Sakai-san (commented):
@sanmiawi @justmejulian Let s have the `state` within the `Modal.examples.tsx`.

So this task turns out of having a bigger scope than expected. But let's go for a `well designed` solution. So let's close the present PR. Thanks for your inputs.

PR #11, opened 2021-04-07
Update npm dependencies, most notably webpack 4 > 5, remove babel-plugin-react-css-modules

neonnoon:
This updates most but not all npm dependencies (no major bumps in prod dependencies).
The webpack config has been updated to work with webpack 5.
Some dependencies have been cleaned up (removed if deprecated an unnecessary, moved from dev to prod or vice versa).

All done for `pyrene`, `pyrene-graphs`, `tuktuktwo` and `kitchensink`.

- [x] depends on #7 
- [x] depends on #108 (otherwise styles will be broken)
- [x] verify styles (currently broken due to the abvove)
- [x] manually verify most important functionality (or introduce <https://www.cypress.io>?)

PR #127, opened 2021-05-31
Update eslint parser to new package

ChrisJamesC:
babel-eslint is deprecated.

### DIFF
--- a/eslint-config/index.js
+++ b/eslint-config/index.js
@@ -98,7 +98,7 @@ module.exports = {
     "mount": "true",
   },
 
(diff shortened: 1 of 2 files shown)


neonnoon:
Nice catch!

While this change of course is safe, I have prepared package updates in #129 (which is based on #11 (which basically touches almost every single package version)). If we can get #11 merged soonish, I'd recommend that we go for #129 to prevent merge conflicts. If it cannot wait or #11 is blocked for much longer, feel free to go ahead here of course.

neonnoon:
Merged as part of #129

PR #144, opened 2021-07-15
Fix BarChartTable

sanmiawi:
This should fix the display problems in the BarChartTable component by updating pyrene being used in pyrene-graphs


Sakai-san (approved):

sosostris (approved):
Ah! So it's only about updating the version of pyrene! nice !!! :D :D :D

Sakai-san (approved):
Thanks a lot !!!!! @sanmiawi

PR #151, opened 2021-08-17
Improve multi value labels

cyrilwanner:
This implements the following changes in the `MultiSelect` component:
- `title` attribute is added to all value labels (so hovering over a truncated label will show the full one)
- the max width of a value label can be configured (previously hardcoded to `123px`)

Please let me know if you prefer a different API for this component or if the `title` attribute should not be added by default.

### DIFF
--- a/pyrene/src/components/MultiSelect/MultiSelect.jsx
+++ b/pyrene/src/components/MultiSelect/MultiSelect.jsx
@@ -197,7 +197,7 @@ MultiSelect.defaultProps = {
   helperLabel: '',
   invalidLabel: '',
   title: '',
-  maxValueLabelWidth: 123,
+  maxValueLabelWidth: '123px',
   name: '',
   defaultValue: [],
   options: [],
@@ -261,7 +261,7 @@ MultiSelect.propTypes = {
   /**
    * Maximum width of the value label in pixels. Use false to disable a maximum width.
    */
-  maxValueLabelWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf([false])]), // eslint-disable-line react/no-unused-prop-types
+  maxValueLabelWidth: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
   /**
    * Sets the html name property of the form element.
    */
--- a/pyrene/src/components/MultiSelect/MultiSelect.jsx
+++ b/pyrene/src/components/MultiSelect/MultiSelect.jsx
@@ -189,6 +189,7 @@ MultiSelect.defaultProps = {
   helperLabel: '',
   invalidLabel: '',
   title: '',
+  maxValueLabelWidth: 123,
   name: '',
   defaultValue: [],
   options: [],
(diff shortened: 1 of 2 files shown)


pyrene/src/components/MultiSelect/MultiSelect.jsx
```diff
@@ -249,6 +258,10 @@ MultiSelect.propTypes = {
    * Displays a loading indicator inside of the input.
    */
   loading: PropTypes.bool,
+  /**
+   * Maximum width of the value label in pixels. Use false to disable a maximum width.
+   */
+  maxValueLabelWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf([false])]), // eslint-disable-line react/no-unused-prop-types
```
justmejulian, line 264: I think we should just stick with the normal css values.
none, 10%, 42em, max-content ...

That would also remove the need for false. -> none

https://developer.mozilla.org/en-US/docs/Web/CSS/max-width

cyrilwanner: Thank you for the feedback, that definitely makes sense! I'll update it.

justmejulian (approved):

sosostris:
The size is totally good! But I'm not so sure about using the title attribute. It seems there's a long delay before it shows it. I tried it in Firefox and Chrome, looks like in Firefox it's a bit faster but in Chrome it's a second, see the gif from kitchensink.

Aug-24-2021 10-05-58

@cyrilwanner have you looked into using the Tooltip component instead of using the title attribute?

sosostris:
Also somehow the icons are not centrally aligned with the text it seems ... 🤔  Could you please also fix it (in a separate PR or just this one both are fine since it should be quite small 😄 )

Screenshot 2021-08-24 at 10 14 42

cyrilwanner:
I am sorry, I should have been more clear on what I've changed since the name of the subcomponents of the `react-select` library are not that self-explanatory (e.g. `MultiValueLabel`).

I changed the following component, so once you select a value and it goes to the list above. I didn't touch the list below the input field with all possible values.

image

> @cyrilwanner have you looked into using the Tooltip component instead of using the title attribute?

I did, but I saw that the `title` attribute was already used for other subcomponents within this `MultiSelect` component (for example within the list from your screenshare), so I went with that to be consistent. It is probably also better regarding accessibility to use standardized browser features.

sosostris:
> I am sorry, I should have been more clear on what I've changed since the name of the subcomponents of the `react-select` library are not that self-explanatory (e.g. `MultiValueLabel`).
> 
> I changed the following component, so once you select a value and it goes to the list above. I didn't touch the list below the input field with all possible values.
> 
> image
> 
> > @cyrilwanner have you looked into using the Tooltip component instead of using the title attribute?
> 
> I did, but I saw that the `title` attribute was already used for other subcomponents within this `MultiSelect` component (for example within the list from your screenshare), so I went with that to be consistent. It is probably also better regarding accessibility to use standardized browser features.

Ah! Thanks a lot for the clarification! Now I get it! I guess since the responsive time of title was so long, I was never aware that we had it for the dropdown options 😂  so I thought this title thing was complete new feature for the multiselect. But yeah, if we're already using it let's just go with it :)

sosostris (approved):

PR #152, opened 2021-08-23
Allow modal with scrolling content to take full window height

cyrilwanner:
This allows a modal with scrolling content to take the full window height instead of the hardcoded max `512px`. So instead of always displaying the scrollbars (even if there is enough space left on the screen) it will now use this available space.

If the content is too large for the screen, it will fall back to display the scrollbar.

To make it backwards compatible and not change all existing modals, it is introduced as an additional property.

### DIFF
--- a/pyrene/src/components/Modal/modal.css
+++ b/pyrene/src/components/Modal/modal.css
@@ -87,7 +87,7 @@
   }
 
   &.contentFullHeight {
-    /* window height - header - footer - 24px padding on both sides */
+    /* window height - header - footer - 24px spacing on both sides */
     max-height: calc(100vh - 64px - 64px - 48px);
   }
 
--- a/pyrene/src/components/Modal/modal.css
+++ b/pyrene/src/components/Modal/modal.css
@@ -86,6 +86,11 @@
     max-height: 512px;
   }
 
+  &.contentFullHeight {
+    /* window height - header - footer - 24px padding on both sides */
+    max-height: calc(100vh - 64px - 64px - 48px);
+  }
+
   &.overlay {
     opacity: 0.5;
     pointer-events: none;
--- a/pyrene/src/components/Modal/modal.css
+++ b/pyrene/src/components/Modal/modal.css
@@ -9,15 +9,21 @@
   left: 0;
   right: 0;
   bottom: 0;
+  padding: 24px;
   background-color: rgba(29, 39, 59, 0.5);
   z-index: 2;
 
   display: flex;
   align-items: center;
   justify-content: center;
+  box-sizing: border-box;
 }
 
 .modalContainer {
+  display: flex;
+  flex-direction: column;
+  max-height: 100%;
+  overflow: auto;
   box-sizing: border-box;
 
   background-color: var(--background-light);
@@ -68,6 +74,7 @@
   box-sizing: border-box;
 
   &.contentScrolling {
+    overflow: auto;
     border-top: solid 1px var(--border);
     border-bottom: solid 1px var(--border);
   }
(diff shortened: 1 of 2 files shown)


pyrene/src/components/Modal/modal.css
```diff
@@ -86,6 +86,11 @@
     max-height: 512px;
   }
 
+  &.contentFullHeight {
+    /* window height - header - footer - 24px spacing on both sides */
+    max-height: calc(100vh - 64px - 64px - 48px);
```
justmejulian, line 91: Oh that just feels like asking for something to break 😬

justmejulian (commented):
What was the specific reason for this change?

Not a big fan of adding additional parameters. IMO either this is how the modal should work or it is not.

Was this looked at with Ivo?

cyrilwanner:
> Was this looked at with Ivo?

Yes, this has been discussed with Ivo and is the preferred behavior according to him.

> What was the specific reason for this change?

In the self-service part, there are some big modals, for example the following one:

<details>
<summary>Click for example</summary>

image
</details>

As the height is currently hardcoded to 512px, you would have to scroll a lot in this modal and you sometimes can't even see the domains on the screens you have added without scrolling.
I've got a ticket to both increase the width & height of the modal to use more of the available screen space. The screenshot above already includes my changes, so it is already wider and you don't see the scrollbar there anymore.

I also can't use the modal without `contentScrolling` since that is not responsive at all and just cuts of the modal if the screen is too small.

<details>
<summary>Click for example</summary>

image
</details>

This prop would actually solve both issues at once: allow bigger heights while still being responsive and be able to display all content on small screens.

---

But please let me know if you have a better idea to achieve a modal with an increased height without having to touch pyrene.

justmejulian:
> Yes, this has been discussed with Ivo and is the preferred behavior according to him.

Oke cool, so then we don't need the extra prop, but just make this the new default?

cyrilwanner:
As discussed, it is now the default behavior and it has been modified to use a flexbox layout.

Sakai-san (approved):
LGTM. Thanks @cyrilwanner !

pyrene/src/components/Modal/modal.css
```diff
@@ -9,15 +9,21 @@
...
   right: 0;
   bottom: 0;
+  padding: 24px;
   background-color: rgba(29, 39, 59, 0.5);
   z-index: 2;
 
   display: flex;
   align-items: center;
   justify-content: center;
+  box-sizing: border-box;
```
Sakai-san, line 19: Don't we have box-sizing by default ?

`box-sizing: border-box;`

Sakai-san (approved):

pyrene/src/components/Modal/modal.css
cyrilwanner: It appears not to be the case, at least not within Kitchensink. Maybe the website that uses this component has set it globally (for example on `*`), but we can't depend on that I think.

justmejulian (approved):

sosostris:
The size is totally good! But I'm not so sure about using the title attribute. It seems there's a long delay before it shows it. I tried it in Firefox and Chrome, looks like in Firefox it's a bit faster but in Chrome it's a second, see the gif from kitchensink.

Aug-24-2021 10-05-58

@cyrilwanner have you looked into using the Tooltip component instead of using the title attribute?

sosostris:
Please also fix the `CONTENT` so that it only shows once in kitchensink demo modal 😆

sosostris:
Thanks for the fixing :) it is good 👍

PR #157, opened 2021-08-30
storybook story button

Sakai-san:


sosostris (approved):
@Sakai-san Thanks a lot for taking the `IconNames` part out of the Storybook POC PR 👍 

I was more thinking it would be nice if we could make a separate PR to handle that part alone targeting for `main` branch; then another separate PR targeting for `main` to handle the module conversion of the css.

Then we could rebase the branch and the Storybook PR would be super compact, containing only the new configs and the nice little example :) In that way we keep things on small scale and concerns separate :)

But this of course also works! Thanks a lot for the effort!

PR #158, opened 2021-08-31
Tooltip story

Sakai-san:


pyrene/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { Story, Meta } from '@storybook/react';
+
+import * as ToolTipComponent from './Tooltip';
```
cyrilwanner, line 5: I am not aware how you are usually doing it in other components. But do you prefer using the star import and then using it like `ToolTipComponent.TooltipProps` instead of `import ToolTipComponent, { TooltipProps } from './Tooltip'`?

cyrilwanner (approved):
Looks good to me! Just added a comment to clarify something, so I will also do the same for further stories.

PR #159, opened 2021-08-31
tabview story

Sakai-san:
### Note
I tried to convert the component into a `Functional Component`. I could not unfortunately due to the actual design of the component.


pyrene/src/components/TabView/TabView.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+);
+
+type ProviderState = {
+  tabName?: number | undefined;
+};
+
+const Template: Story<TabViewProps> = () => (
+  <StateProvider<TabViewProps, ProviderState> initState={{ tabName: undefined }}>
+    {
+      (state, setState) => ({
```
cyrilwanner, line 27: Did you consider changing the API of `StateProvider` to directly return a react component here instead of an object? I think that would be more react-like and also allow to return nested components (maybe this is needed for another story?) or basically anything that is valid JSX.

For example:
```jsx
(state, setState) => (
  <TabViewComponent
    initialTabName="Tab 1"
    directAccessTabs={3}
    ...
  />
)
```

cyrilwanner (commented):
It looks very nice!
But somehow, I cannot modify the props of the TabView component. Can we somehow fix this? It was working with Kitchensink so I think we should try to also make it work with Storybook, otherwise it would be a downgrade for this component 😅

Sakai-san:
> It looks very nice!
> But somehow, I cannot modify the props of the TabView component. Can we somehow fix this? It was working with Kitchensink so I think we should try to also make it work with Storybook, otherwise it would be a downgrade for this component 😅

You are right. I need to fix that. Maybe it because I did not use this

```
Tabview.arg = {
  initialTabName="Tab 1"
  ...
}
```

Sakai-san:
> Thank you very much for the changes! It looks great and all Storybook features are working now 🎉

Thanks a lot for your inputs @cyrilwanner . Krasse Teamwork.

PR #169, opened 2021-09-14
Cleanup

Sakai-san:
1. Use` class properties`, like at [Facebook](https://github.com/facebook/react/issues/9851), so no need to `bind` context in methods.
2. `Reuse` Type.
3. Remove useless `React.Fragment`
4. Enhanced `code readability`
5. Use chaining operator with [function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls)
6. Enhance `check`
7. Remove useless bracket `{` in function


cyrilwanner (approved):
Thank you for cleaning those things up, that's definitely better this way!

PR #171, opened 2021-09-15
Fix issues with Pyrene color page.

sanmiawi:
Dod:
Remove border from chart 3
Add color definition for Semantic Scheme

Screenshot 2021-09-13 at 13 24 18
Screenshot 2021-09-13 at 13 26 48

### DIFF
--- a/kitchensink/src/app/static/foundations/ColorsPage.jsx
+++ b/kitchensink/src/app/static/foundations/ColorsPage.jsx
@@ -92,7 +92,7 @@ const ColorsPage = () => (
           description=""
           rowHeader={{
             colors: [{
-              colorName: 'Information', name: 'blue-500', darkFont: false, bordered: false,
+              colorName: 'Information', name: 'blue-500', darkFont: false, bordered: false, infoText: 'blue-500',
             }],
             width: '200px',
           }}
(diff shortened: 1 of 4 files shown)


kitchensink/src/app/static/foundations/ColorsPage.jsx
```diff
@@ -92,7 +92,7 @@ const ColorsPage = () => (
           description=""
           rowHeader={{
             colors: [{
-              colorName: 'Information', name: 'blue-500', darkFont: false, bordered: false,
+              colorName: 'Information', name: 'blue-500', darkFont: false, bordered: false, infoText: 'blue-500',
```
justmejulian, line 95: what is the difference to name?

justmejulian (changes requested):

kitchensink/src/app/static/foundations/ColorsPage.jsx
justmejulian, line 95: Why? what is the difference to name?

sanmiawi: name is a variable and the colorbox cannot be displayed without it. infoText contains text that describes the color and it can be omitted.

justmejulian:
Thanks

justmejulian (approved):

PR #173, opened 2021-09-15
form ts [feature]

Sakai-san:


pyrene/src/components/Form/Form.tsx
```diff
@@ -0,0 +1,258 @@
...
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import { ValidationError, ObjectSchema } from 'yup';
+import { MultiselectOption, Filters, Options } from '../Filter/types';
+
+type ValuesOf<T> = T[keyof T];
+type KeysOf<T> = keyof T;
+
+type FormValues = Filters;
+type ToucheValues = {
```
maximelovino, line 14: I would rename this type to `TouchedValues`

pyrene/src/components/Form/Form.tsx
```diff
@@ -0,0 +1,258 @@
...
+type InputComponentProps = {
+  name: KeysOf<FormState['values']>,
+  value: ValuesOf<FormState['values']>,
+  invalid: boolean,
+  invalidLabel: string,
+  onChange: (value: string, event: React.ChangeEvent<any>) => void,
+  onBlur: (event: React.FocusEvent<any>) => void,
+};
+
+const getTouchedState = (initialValues: FormValues) => (
```
maximelovino, line 52: Since this is only for computing the initial state, I would rename it to `getInitialTouchedState`

maximelovino (changes requested):
The type and the function should be renamed imo

pyrene/src/components/Form/Form.tsx
Sakai-san: @maximelovino Ok. I ll do so. This task was originally only to do the migration. But I can also do the renaming.

pyrene/src/components/Form/Form.tsx
Sakai-san: > I would rename this type to TouchedValues
Thanks. Nice catch !

maximelovino (approved):

PR #174, opened 2021-09-17
multiselect example [improvement]

Sakai-san:


cyrilwanner (approved):
That looks way better, thanks!

PR #175, opened 2021-09-17
multiselect utils TS [improvement]

Sakai-san:


pyrene/src/components/MultiSelect/delimiterUtil.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
cyrilwanner, line 1: What is the reason to disable this rule?

Sakai-san: > What is the reason to disable this rule?

Hello @cyrilwanner. Yes the `linter` complains about the return type. I most of the time let `tsc` to `infer` type and do not hardcode the return type myself.

Sakai-san: Error message by the linter:

> Missing return type on function.eslint@typescript-eslint/explicit-module-boundary-types

cyrilwanner: Okay thanks, if that is the preferred way, it makes sense 👍 But it might be worth updating it in the OSAG eslint package so you don't have to add it to all files (but not in this PR :)).

cyrilwanner (approved):

PR #176, opened 2021-09-20
multiselelect ts [improvement]

Sakai-san:


pyrene/src/components/MultiSelect/types.ts
```diff
@@ -1,5 +1,8 @@
+import { IconProps } from '../Icon/Icon';
+
 export interface Option {
+  iconProps?: IconProps,
   label: string,
-  invalid: boolean,
-  value: string | number,
+  value: string,
```
sanmiawi, line 6: Is there a reason why `number` is not accepted as a possible type for value?

sanmiawi (approved):

pyrene/src/components/MultiSelect/types.ts
Sakai-san: Yes, this is due to the `react-select` typing (npm dependency).

PR #178, opened 2021-09-22
filter ts [improvement]

Sakai-san:


pyrene/src/components/Filter/FilterComponents/FilterBar.tsx
```diff
@@ -1,9 +1,44 @@
...
+   * */
+  filterValues: FilterValues,
+  /**
+   * True to enable the visual components to handle negated filters.
+   */
+  negatable: boolean,
+  /**
+   * Called when the user clicks on the apply button. Exposes two parameters: filterValues and negatedFilterKeys (contains an array of the keys of the filters that are negated).
+   */
+  onFilterSubmit?: (filterValues: FilterBarProps['filterValues'], negatedKeys: Array<Filter['id']>) => void,
```
justmejulian, line 32: what is the benefit of FilterBarProps['filterValues'] over just using FilterValues?

pyrene/src/components/Filter/FilterComponents/FilterBar.tsx
```diff
@@ -28,9 +63,15 @@ import FilterTag from './FilterTag';
...
  */
 
-export default class FilterBar extends React.Component {
+export default class FilterBar extends React.Component<FilterBarProps, FilterBarState> {
+
+  static displayName = 'FilterBar';
 
-  constructor(props) {
+  static defaultProps = {
+    onFilterSubmit: () => null,
```
justmejulian, line 71: I prefer undefined over empty function calls, empty function calls are still called.
And I personally find it more readable to know that a function does not exist.

justmejulian (approved):

PR #180, opened 2021-09-24
filter-cleanup [improvement]

Sakai-san:


pyrene/src/components/SingleSelect/SingleSelect.tsx
```diff
@@ -169,94 +172,52 @@ const SingleSelect = <ValueType extends unknown = DefaultValueType>({
...
   value,
-  onChange = () => null,
+  onChange,
   onBlur = () => null,
   onFocus = () => null,
 }: SingleSelectProps<ValueType>): React.ReactElement => {
 
   const optionsObj = getOptionsObj(options, groupedOptions, sorted);
 
+  const props = {
```
cyrilwanner, line 182: Maybe we should call this `selectProps` (or something else)? Normally, I think `props` is reserved for the component props itself that are passed in as a function argument to not cause confusion when used later in the file.

cyrilwanner (approved):
Looks good to me, just one very small wording suggestion

PR #182, opened 2021-09-28
code cleanup [improvement]

Sakai-san:
# Done in that PR

1. Let the TS compiler do `type inference` on its own.
2. Fix bug happening while `clearing up` the input field filter introduced in [PR-176](https://github.com/open-ch/pyrene/pull/176/files) (`onChange`)
3. Stop unnecessary `call of void` `event handler`.
4. Remove useless `type of parameter` in FunctionComponent for non-public component.
5. `Nullish` operator
6. Remove useless `parenthesis`

### DIFF
--- a/pyrene/src/components/CheckboxPopover/CheckboxList.tsx
+++ b/pyrene/src/components/CheckboxPopover/CheckboxList.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/prop-types */
 import React, { FunctionComponent } from 'react';
 
 import Checkbox from '../Checkbox/Checkbox';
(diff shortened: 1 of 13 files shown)


pyrene/src/components/CheckboxPopover/CheckboxList.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/prop-types */
```
cyrilwanner, line 1: Should we disable this globally (eslintrc) instead of in every file?

cyrilwanner (approved):
Looks good, just one suggestion as I would prefer to not always have to disable the same eslint rule in every TS file.

pyrene/src/components/CheckboxPopover/CheckboxList.tsx
Sakai-san: We still need this rule for making sure the `public Pyrene components` work well.

For these component we have to `type` the props of the function component. If we do not do that the `babel plugin` cannot generate the `proptype` for `kitchensink`,

PR #184, opened 2021-09-29
radiopopover ts [improvement]

Sakai-san:


pyrene/src/components/RadioPopover/RadioPopover.tsx
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable react/require-default-props */
+import React, { FunctionComponent, useState } from 'react';
+import clsx from 'clsx';
+
+import styles from './radioPopover.css';
+import Popover from '../Popover/Popover';
+import OptionList from './OptionList';
+import { Option } from './types';
+
+export interface RadioPopoverProps {
```
cyrilwanner, line 10: I think `onChange`, `options` and `renderHelpSection` are only used for `<OptionList>`. Can we somehow avoid the duplicate type definitions by extending them? Either here or in OptionList?
If it is too much work or you think it is not worth it, ignore it :)

cyrilwanner (approved):

PR #185, opened 2021-09-29
Set up Typescript for pyrene-graphs and tuktuktwo

sosostris:
This PR sets up Typescript in the same way as in Pyrene, which is prerequisite to using Storybook for the visualization components.

Based on the last discussion with designer, we're going to have a "Visualizations" section parallel to the "Components" section in Storybook, where all visiaulization-related components (all components in tuktuktwo and pyrene-graphs) will be displayed, and no nested structure.

Since the current kitchensink is not displayed tuktuktwo components anyway, the hack for the `docgen` and `typescript-to-proptypes` hack is not added.


pyrene-graphs/jest.config.js
```diff
@@ -168,6 +168,8 @@ module.exports = {
   transform: {
     '^.+\\.js$': 'babel-jest',
     '^.+\\.jsx?$': 'babel-jest',
+    '^.+\\.ts$': 'babel-jest',
+    '^.+\\.tsx?$': 'babel-jest',
```
cyrilwanner, line 172: I think the second line already covers the first one (`x?` -> x is optional and it matches both `.ts` and `.tsx`). Same for the js part.

tuktuktwo/jest.config.js
```diff
@@ -168,6 +168,8 @@ module.exports = {
   transform: {
     '^.+\\.js$': 'babel-jest',
     '^.+\\.jsx?$': 'babel-jest',
+    '^.+\\.ts$': 'babel-jest',
+    '^.+\\.tsx?$': 'babel-jest',
```
cyrilwanner, line 172: Same here

cyrilwanner (approved):
Thank you, it looks good to me!
Just one suggestion to simplify a config file, it you want to apply this :)

Sakai-san (approved):
Thanks a lot !!

pyrene-graphs/jest.config.js
sosostris: I totally agree! lol ! Thanks for the sharp eye :) Let me clean this up!

PR #186, opened 2021-09-29
storybook repo upper [improvement]

Sakai-san:


Sakai-san:
> Thank you for regenerating the lock file, installation works now also outside of our VPN. The rest also looks good 👍

@cyrilwanner Thanks. There is something wrong. I think some npm dependency is needed for storybook to do the props parsing. see before and now. I am still investigating.

Before
Screenshot 2021-10-07 at 14 12 56
After

Screenshot 2021-10-07 at 15 20 57

PR #188, opened 2021-09-30
radiopopover ts [improvement]

Sakai-san:
# Done in that PR
1. Revert back the `extends` of props' type of `OptionListProps` done in [PR-184](https://github.com/open-ch/pyrene/pull/184). **Babel** plugin does not work with `algebraic data type`.
2. Component Demo for `Kitchensink` typing.

Screenshot 2021-09-30 at 14 30 16
Screenshot 2021-09-30 at 14 30 49


cyrilwanner (approved):
Sorry, I forgot that the babel plugin for Kitchensink will fail 😅 
So we can only improve it when everything is migrated to Storybook.

PR #189, opened 2021-09-30
searchfinder ts [improvement]

Sakai-san:


pyrene/src/components/SearchFinder/SearchFinder.tsx
```diff
@@ -1,23 +1,50 @@
...
+const SearchFinder: FunctionComponent<SearchFinderProps> = ({
   onSearchTermChange,
   searchTerm,
   resultCount,
   selectedResult,
-  placeholder,
+  placeholder = '',
   onSelectedResultChange,
-}) => {
+}: SearchFinderProps) => {
```
cyrilwanner, line 47: Do you want to remove `SearchFinderProps` here like you did in other PRs?

cyrilwanner (approved):

pyrene/src/components/SearchFinder/SearchFinder.tsx
Sakai-san: @cyrilwanner Here I need to keep it. Because it s an **exposed** component (not internal). Hence the `babel plugin` has to generate the `PropType` for the documentation system (`kitchensink`).

cyrilwanner: Ah, so we can only remove it on non-exposed components for now. Cool, thank you for clarifying!

PR #215, opened 2021-11-03
typescript migration [improvement]

Sakai-san:
# Done in the present PR

- Fix `bug` introduced by that [PR-212](https://github.com/open-ch/pyrene/pull/212/files) (due to default props removal)
- TypeScript `migration` completed for the `TreeTable`, namely `example` for kitchensink
- Code simplified


pyrene/src/components/TreeTable/TreeTable.tsx
```diff
@@ -26,7 +28,8 @@ export interface TreeTableProps<R>{
    * Sets the Table data displayed in the rows.
    * Type: [{ children: object, lineCount: number, ...row }]
    */
-  data: Array<RowData<R>>,
+  data: Array<R>,
+  defaultExpandedSection?: string,
```
cyrilwanner, line 32: I think this is a public component, so can we also have a description/comment for this prop?

cyrilwanner (approved):

pyrene/src/components/TreeTable/TreeTable.tsx
Sakai-san: > I think this is a public component, so can we also have a description/comment for this prop?

The props `defaultExpandedSection` is actually not used in the `TreeTable` component. So I just removed it. Thanks @cyrilwanner

PR #218, opened 2021-11-08
DateTimePicker Component

sanmiawi:
This is a component for selecting a date via a dropdown calendar or entering it using a text input.

It integrates a Pyrene styled input component with react-datepicker library.
Pyrene styled input provides some validation features and the react-datepicker library provides the dropdown calendar.

For some history on the journey so far, please check the [POC PR](https://github.com/open-ch/pyrene/pull/177)
For OSAG employees there is an internal JIRA issue tracking the development of this component with mockups and links to research pages. Please search in JIRA or reach out to me for the link.


dparret:
Thanks for the PR @sanmiawi!

pyrene/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Checks if a timetype is valid
+ * @param {TimeType} time
+ */
+export const isValidTime = (time?: TimeType): boolean => {
+  if (time && time.hours >= 0 && time.hours <= 23 && time.minutes >= 0 && time.minutes <= 59) {
+    return true;
+  }
+  return false;
```
maximelovino, line 100: Here you can just return:

```
return time && time.hours >= 0 && time.hours <= 23 && time.minutes >= 0 && time.minutes <= 59
```

pyrene/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Checks if a timetype is valid
+ * @param {DateType} date
+ */
+export const isValidDate = (date?: DateType): boolean => {
+  if (date) {
+    return isExists(date.year, date.month - 1, date.day);
+  }
+  return false;
```
maximelovino, line 89: Here you can just return:

```
return date && isExists(date.year, date.month - 1, date.day)
```

maximelovino (commented):
In the input it shows 24h times and in the date picker it shows 24h time but with also the AM/PM behind which shouldn't be the case no?

image
Also if i click on a time before clicking on a date, it just selects the time and puts only time in the input:
https://user-images.githubusercontent.com/17678807/140884004-4f3bb301-271e-4cbc-b1c5-008c5e386503.mov

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+  errorValue,
+  disabled,
+  label = dateOnly ? 'Date' : 'Date & Time',
+  name,
+  onChange,
+  onClick,
+  onFocus,
+  setDateValue,
+  setTimeValue,
+  timeFormat = ' HH:mm',
```
Sakai-san, line 81: The `formatting` of data is something that should be separated to `data` itself.

Like presentation (HTML) and style (CSS) if you want.

emandilara:
A point related to the UI/UX of the DateTimePicker:

I think it would be better to not show the AM, PM abbreviations if we keep the 24-hour clock convention (i.e. I'd rather have 15.30 or 3.30 PM). Also, if we keep the abbreviations, I'd suggest to show them in the form input together with the date for consistency (e.g. 12.11.2021 3:00 PM). As there is some space left, would it add some value to show the timezone there too?

@ibaldi what do you think?
image

Sakai-san:
The first time you click in the time on the right hand side the date does not populate.

https://user-images.githubusercontent.com/6510794/140885545-c8477428-f140-4325-9947-36b73f3c74ca.mov

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
```diff
@@ -0,0 +1,138 @@
+
```
Sakai-san, line 1: Do you need this line on the top ?

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
```diff
@@ -0,0 +1,194 @@
...
+  customInput = <input />,
+  inline = false,
+  isOpen,
+  onCalendarOpen,
+  onChange = () => {},
+  onClickOutside,
+  onSelect,
+  openDate,
+  range = false,
+  timeFormat = ' HH:mm',
```
Sakai-san, line 115: The `formatting` of data is something that should be separated to `data` itself.

Like presentation (HTML) and style (CSS) if you want.

pyrene/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,227 @@
+import isExists from 'date-fns/isExists';
+import sub from 'date-fns/sub';
+import add from 'date-fns/add';
+import format from 'date-fns/format';
+import parse from 'date-fns/parse';
+import Duration from 'date-fns';
+
+import {
```
Sakai-san, line 8: Same as above.
```
import zonedTimeToUtc from 'date-fns-tz/zonedTimeToUtc';
import utcToZonedTime from 'date-fns-tz/utcToZonedTime';
```

[Performance](https://stackoverflow.com/questions/35250500/correct-way-to-import-lodash) things.

kitchensink/src/app/static/howToPages/DateTimePicker.jsx
```diff
@@ -0,0 +1,151 @@
...
+  maxDateTime: 10236782323493,
+  timeZone: 'Europe/Zurich',
+};
+
+DateUsage.propTypes = {
+  dateOnly: PropTypes.bool,
+  maxDateTime: PropTypes.number,
+  minDateTime: PropTypes.number,
+  timeZone: PropTypes.string,
+};
```
Sakai-san, line 143: `DateTimePicker` is written in `TypeScript` right ? I do not understand why we have this `propTypes` here.

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+   * Time format used by component
+   */
+  timeFormat?: string,
+  /**
+   * Input time value
+  */
+  timeValue?: string,
+}
+
+const DateTimeInput = forwardRef(({
```
Sakai-san, line 68: `forwardRef<HTMLInputElement, DateTimeInputProps>`

dparret:
Thanks for the PR @sanmiawi! I'm looking forward to this component. :)
I will only review the UX/Design angle (I'm following Ivo's mock and not sure what was exactly specified).

1. Default state.
- The default state should be with no preselected date. 
image vs 
image
- And when playing with the month, it seems the preselected date (that shouldn't be there) change.
https://user-images.githubusercontent.com/67573484/140887390-3fbe213e-8ca7-417b-be7e-ffdd96ed1044.mov

2. The above would also probably prevent having an empty date if the time only is selected.
https://user-images.githubusercontent.com/67573484/140881678-f0ccf943-600e-48dd-a9ea-3cab79caa535.mov

3. Max and min dates.
- Dates from 1 year+ are gray and they can't be selected by default. I would prefer having no maximum as default.
image
- I feel that the maxDateTime doesn't do what it should (at least the example in kitchensink). 
image I couldn't build an example with minDateTime as well.

4. Time format.
I don't know what was specified regarding AM/PM, so disregard if I'm wrong.
- PM hours should be below 12, 22:30 PM doesn't exist.
image
- We should be able to manually type in AM or PM if we use that format in the dropdown.
image
- AM/PM is not displayed in the text box when we select an hour, but is in the dropdown.
image
It should probably look like this 
image.
The 2 points above lead me to think that the time format is not properly used in the dropdown (displaying AM/PM should be decided with the time format and used in the 2 places - dropdown and input field).

5. If we choose to have a date and time, the time should always be displayed. Even if 0:00.
image

6. The closeOnSelect enables a user to have 2 dates selected. Not sure what's happening, but we definitely don't want two dates to be highlighted.
image

7. I'm definitely unsure about that point as I don't know how it was specified. In terms of UX, it's strange that a user can enter a valid date that isn't highlighted in the dropdown. If the user specifies a time manually, it has to be somehow reflected in the dropdown (or the input should not be valid).
image

8. Date only still displays the hour. 
image

9. Again, not sure about the spec, but what's the purpose of the inline props? Do we already have a use case for such?

10. A change in the timeFormat removes the space between the date and the time. 
image

11. What's the purpose of the endDate props? Same question regarding the startDate, selectDate, selectTime props. I actually don't expect to have anything related to a range in a date & time picker. We should also probably remove the example at the bottom, as it shouldn't be part of the component.

12. Not sure how the timeStamp props work, but when playing with it, the picker is stuck in 1970. The props doesn't seem to change anything. 
image

Now, let's go in nitpicking mode...
13. The month is not vertically centered with the 2 arrows. 
image

14. When switching from shorter months (only 5 weeks and not 6) to longer months (6 weeks). The overall size increases. 
https://user-images.githubusercontent.com/67573484/140888604-d8c63a5a-1293-41c0-9e0d-3176ab797b11.mov

15. If the default value of closeOnSelect is true, the box should be toggled. 
image

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
```diff
@@ -0,0 +1,194 @@
...
+   */
+  minDate?: Date,
+  /**
+   * Do something when calendar opens
+   */
+  onCalendarOpen?: () => void,
+  /**
+   * Function to handle date change event
+   */
+  onChange?: (date: Date | [Date, Date] | null, event: React.SyntheticEvent<any> | undefined) => void,
```
Sakai-san, line 52: Try to reuse type instead of copying it :

`onChange?: ReactDatePickerProps['onChange']`

kitchensink/src/app/static/howToPages/DateTimePicker.jsx
sanmiawi: This is the HowToPage in Kitchensink. There is not tsx in Kitchensink

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
sanmiawi: Copy and paste🤦🏾‍♂️

sanmiawi:
> Thanks for the PR @sanmiawi! I'm looking forward to this component. :) I will only review the UX/Design angle (I'm following Ivo's mock and not sure what was exactly specified).
> 
> 1. Default state.
> 
> * The default state should be with no preselected date.
>   image vs
>   image
> * And when playing with the month, it seems the preselected date (that shouldn't be there) change.
>   https://user-images.githubusercontent.com/67573484/140887390-3fbe213e-8ca7-417b-be7e-ffdd96ed1044.mov
> 
> 1. The above would also probably prevent having an empty date if the time only is selected.
>    https://user-images.githubusercontent.com/67573484/140881678-f0ccf943-600e-48dd-a9ea-3cab79caa535.mov
> 2. Max and min dates.
> 
> * Dates from 1 year+ are gray and they can't be selected by default. I would prefer having no maximum as default.
>   image
> * I feel that the maxDateTime doesn't do what it should (at least the example in kitchensink).
>   image I couldn't build an example with minDateTime as well.
> 
> 1. Time format.
>    I don't know what was specified regarding AM/PM, so disregard if I'm wrong.
> 
> * PM hours should be below 12, 22:30 PM doesn't exist.
>   image
> * We should be able to manually type in AM or PM if we use that format in the dropdown.
>   image
> * AM/PM is not displayed in the text box when we select an hour, but is in the dropdown.
>   image
>   It should probably look like this
>   image.
>   The 2 points above lead me to think that the time format is not properly used in the dropdown (displaying AM/PM should be decided with the time format and used in the 2 places - dropdown and input field).
> 
> 1. If we choose to have a date and time, the time should always be displayed. Even if 0:00.
>    image
> 2. The closeOnSelect enables a user to have 2 dates selected. Not sure what's happening, but we definitely don't want two dates to be highlighted.
>    image
> 3. I'm definitely unsure about that point as I don't know how it was specified. In terms of UX, it's strange that a user can enter a valid date that isn't highlighted in the dropdown. If the user specifies a time manually, it has to be somehow reflected in the dropdown (or the input should not be valid).
>    image
> 4. Date only still displays the hour.
>    image
> 5. Again, not sure about the spec, but what's the purpose of the inline props? Do we already have a use case for such?
> 6. A change in the timeFormat removes the space between the date and the time.
>    image
> 7. What's the purpose of the endDate props? Same question regarding the startDate, selectDate, selectTime props. I actually don't expect to have anything related to a range in a date & time picker. We should also probably remove the example at the bottom, as it shouldn't be part of the component.
> 8. Not sure how the timeStamp props work, but when playing with it, the picker is stuck in 1970. The props doesn't seem to change anything.
>    image
> 
> Now, let's go in nitpicking mode... 13. The month is not vertically centered with the 2 arrows. image
> 
> 1. When switching from shorter months (only 5 weeks and not 6) to longer months (6 weeks). The overall size increases.
>    https://user-images.githubusercontent.com/67573484/140888604-d8c63a5a-1293-41c0-9e0d-3176ab797b11.mov
> 2. If the default value of closeOnSelect is true, the box should be toggled.
>    image






1. Noted
2. I am not sure that this would prevent an empty date if time is selected first. What I think will prevent an empty date is if we assume that the date is `now` on time select or we prevent time from being selected before a date has been selected.
3.  

- Yes default can be no maximum date
- maxdate does what it should, a timestamp of 6 is in 1970 not 6 years in the future

4. Yes the AM/PM situation is a bit confusing @emandilara has mention this too. I hope we can find the best way to handle this in the meeting today. It does not make sense to have 22:00 PM in the drop down. If you check the Zeplin mockups we do not have this confusion. What we have is currently a mix between Zeplin and Figma. That is why the AM/PM is not shown in the input box because that is how it is mocked in Zeplin.
5. I am not sure if we want to assume the date especially for MC, we  can definitely do this. But because of issues like
[this](https://github.com/date-fns/date-fns/issues/1998) and [this](https://github.com/date-fns/date-fns/issues/1849) we decided that the engineer has to enter the date as expected.
6. This is supposed to toggle the state of being able to close the calendar dropdown on date select. This does not work well with kitchensink. Will look into the double date select.
7. This is true but we are not sure that the selected time will always be in 30 mins intervals.
8. This is because you are changing the dateOnly value after the dateTime has been rendered. we do not hold this value in a state because we do not expect this value to change after the component has been rendered
9. Inline is just a property that is available on react-datepicker library that renders the dropdown calendar on render. We do not need it and can be removed.
10. Yes, this is the result of having a flexible format. If you want a space between the date and time you need to include it in the format.
11. They are all needed for selecting a range. You can check out the From and To example at the bottom of the page.
12. Timestamps are usually very large numbers. so numbers like 0, 1, 2 are usually in 1970 you can check https://www.unixtimestamp.com/ for examples of timestamps
13. Noted
14. I think that is expected, or that is how react-datepicker handles the dropdown because there are more weeks in one than the other. But I can look into fixing it. Can there be a mock for this? Because question will be are we going to show more days from the next month to make up for height or do we just include more padding to increase the heigh?
15. Yes that is true, will look into this but I think it is only checked in kitchensink if the prop is passed to the kitchensink example.

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+  onChange,
+  onClick,
+  onFocus,
+  setDateValue,
+  setTimeValue,
+  timeFormat = ' HH:mm',
+  timeValue = '',
+}:DateTimeInputProps, ref:React.Ref<HTMLInputElement>) => {
+
+  const handleDateOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
sosostris, line 85: The behavior of the input box feels a bit odd:
- inputting `abcd` won't show anything as it doesn't match the regex
- inputting `1234` would be considered valid input but it's apparently not a good time format
- `111111111` is valid but `1111111111` is not valid as this is 1 more digit

large part of this behavior comes from the regex and the logic about `handleDataChange` here, let's discuss about it later :)

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,303 @@
...
+   */
+  name?: string,
+  /**
+   * Calendar opened callback
+   */
+  onCalendarOpen?: () => void,
+  /**
+   * Click outside componet callback
+   */
+  onClickOutside?(event: React.MouseEvent<HTMLDivElement>): void,
```
cyrilwanner, line 66: I think we should type all functions in the same way, so use arrows in this case (do it the same as you did `onCalendarOpen` a few lines above):
`onClickOutside?: (event: React.MouseEvent<HTMLDivElement>) => void,`

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Callback for when user clicks outside component
+   */
+  onClickOutside?(event: React.MouseEvent<HTMLDivElement>): void,
```
cyrilwanner, line 56: Same here

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Function to handle date select event
+   */
+  onSelect?(date: Date, event: React.SyntheticEvent<any> | undefined): void,
```
cyrilwanner, line 60: Same here

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,303 @@
...
+  */
+  selectStart?: boolean,
+  /**
+  * Date object that represents the start date of the component
+  */
+  startDate?: Date,
+  /**
+   * Date format used by component
+   */
+  timeFormat?: string,
```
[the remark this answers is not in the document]
emandilara: The following idea can be combined with my comment about the 12h/24h clock and the `AM`/`PM` abbreviations:
What if we provide an option so that we can choose if we want to have a 12h clock with the abbreviations or a 24h clock without them?

kitchensink/src/app/static/howToPages/DateTimePicker.jsx
```diff
@@ -0,0 +1,151 @@
...
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  );
+};
+
+DateUsage.defaultProps = {
+  dateOnly: false,
+  minDateTime: 0,
```
emandilara, line 133: Can we start with a bigger value here or use a text form to type the value? 🙂
It takes a lot of time to reach a realistic timestamp in Kitchensink by clicking the plus button 😄
Screenshot 2021-11-09 at 13 29 38

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+export interface DateTimePickerProps{
+  /**
+   * Calendar is opened on input component render
+   */
+  calendarOpened?: boolean,
+  /**
+   * Close calendar on date select
+   */
+  closeOnSelect?: boolean,
```
emandilara, line 26: It seems that the `closeOnSelect` does not work as expected; it closes the picker when the `dateOnly` is selected, but if not, the picker stays open.

https://user-images.githubusercontent.com/15834481/140926805-cbcd6f0f-1a26-4890-89b7-ccaf87e7980a.mov

n.

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,303 @@
...
+  /**
+   * Date format used by component
+   */
+  dateFormat?: string,
+  /**
+   * Boolean to toggle time display
+   */
+  dateOnly?: boolean,
```
emandilara, line 34: What if we have a `timeOnly` option as well? I think there could be some use cases where the dates might be pre-decided and only the time should be selected. I would find it very useful to have this functionality wrapped by this component. Of course this feature could be added later, after this PR is closed, I just thought it would be nice to discuss it.

emandilara (changes requested):

sanmiawi:
> Yes. Or we can decide to not close the dropdown when not both (date and time) were selected.

This option can be explored.

> Thanks for the explanation. Let's do no max by default.

👍🏾

> Sorry, not sure I understood your reply here. What I mean is that in case we are in date only, we only display/select a date. But when we have both date and time, we need to display/select both all the time. We shouldn't display a date alone (i.e. with no time) if we are not in the dateOnly case.

I guess if we decide to not close the dropdown when not both (date and time) were selected. Then we will not have this problem anymore

> Thanks, so we can remove it, yes.

👍🏾

> I could suggest two ways here (I don't care much which one), but I find weird to expect the user to start the timeFormat with a space.
> 
> * We merge the date and time format into one.

If we merge date and time format into one we will not be able to have such error displays. We will not be able to tell if it is the date that has the error or the time. We will just have one error, `Date invalid`
Screenshot 2021-11-05 at 11 30 18

> * We had the space ourselves (not the user) in the component when a user is using the timeFormat.

If we want to keep the option to have different errors for date and time then this would be a possible solution. But if we insert a space by ourselves then we are already messing with the format that is available to the user. I guess this is a tradeoff that we should be able to live with.

> It's a date picker and should have nothing to do with a range. If we use 2 date pickers in a range at some point, that wrapper will handle which date picker is the from and which one is the to. It should not be done with that component IMO.

This makes sense but for instance in the Service Windows we are replacing From and To dates. Most of the datepicker libraries that were researched support a range. The range is part of the react-datepicker library and we are just passing props down to it. We are not doing any complicated coding to activate this.

> We can try to stick to the simplest solution here and have a fixed height that handles the 6 rows and leave some space when we only have 5 rows (without changing the dropdown height). Lucky us, 6 rows is the maximum we can have for a month.

Will look into this.

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
emandilara: Thanks for explaining this part in our meeting :) I understood that it needs to have a selected time too before it closes the picker. I am resolving this comment.

kitchensink/src/app/static/howToPages/DateTimePicker.jsx
sanmiawi: From what @dparret said we will be going with no minimum and no maximum defaults.

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
sanmiawi: ✅

pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx
sanmiawi: As discussed we will be keeping the current validation style.

pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx
```diff
@@ -0,0 +1,197 @@
...
+  timeFormat?: string,
+  /**
+   * Should display the Time column on the right-hand side
+   */
+  shouldDisplayTimeColumn?: boolean,
+}
+
+const ReactDatePickerWrapper: React.FC<DatePickerProps> = ({
+  closeDropdown,
+  dateFormat = 'dd.MM.yyyy',
```
Sakai-san, line 94: Maybe you could call the function returning the various format you implemented

 ```
  closeDropdown,
  dateFormat = getFormat( 'eu').dateFormat,
  ....
```

File: pyrene/src/components/DateTimePicker/DateTimePicker.tsx (hunk 11 above)
Comment (sanmiawi) [reply pyrene/src/components/DateTimePicker/DateTimePicker.tsx]: ✅

File: pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx (hunk 12 above)
Comment (sanmiawi) [reply pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx]: ✅

File: pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx (hunk 13 above)
Comment (sanmiawi) [reply pyrene/src/components/DateTimePicker/ReactDatePickerWrapper/ReactDatePickerWrapper.tsx]: ✅

File: pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx (hunk 8 above)
Comment (sanmiawi) [reply pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx]: ✅

File: pyrene/src/utils/DateUtils.ts (hunk 6 above)
Comment (sanmiawi) [reply pyrene/src/utils/DateUtils.ts]: ✅

File: pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx (hunk 4 above)
Comment (sanmiawi) [reply pyrene/src/components/DateTimePicker/DateTimeInput/DateTimeInput.tsx]: ✅

**Comment by dparret** [comment]
We are making good progress here :) 

I've seen still two glitches (in no particular order):
1. In the Date & Time picker, picking a date only seems to be a valid format (no problem with that), but the date shown in the picker doesn't match the one in the field.
https://user-images.githubusercontent.com/67573484/142819335-1edc0e88-03da-4248-bf64-2b24d13e33fb.mov

2. The disabled state doesn't match other components disabled states. 
image
vs
image

**Comment by emandilara** [comment]
I just noticed this strange behaviour. In my opinion, the regular expression works well and shows the "Invalid date format" error but the calendar underneath should not move to the next month.
It's not any big bug, but I think we should keep a note in order to improve the UX during the enhancement phase.

https://user-images.githubusercontent.com/15834481/143005789-2f5fbef8-30c5-4462-960d-ec4f9a93378d.mov

**Comment by dparret** [comment]
> I just noticed this strange behaviour. In my opinion, the regular expression works well and shows the "Invalid date format" error but the calendar underneath should not move to the next month. It's not any big bug, but I think we should keep a note in order to improve the UX during the enhancement phase.
> 
>  Screen.Recording.2021-11-23.at.11.10.17.mov

Nice catch, didn't see that one.

**Review by Sakai-san — CHANGES_REQUESTED** [review]
I found a bug while playing around with the date picker. See the video.

https://user-images.githubusercontent.com/6510794/143009404-a1ac7245-7025-4f6b-ad4a-bcce0e4d0ac4.mov

**Comment by emandilara** [comment]
> We are making good progress here :)
> 
> I've seen still two glitches (in no particular order):
> 
> 1. In the Date & Time picker, picking a date only seems to be a valid format (no problem with that), but the date shown in the picker doesn't match the one in the field.
>    https://user-images.githubusercontent.com/67573484/142819335-1edc0e88-03da-4248-bf64-2b24d13e33fb.mov
> 2. The disabled state doesn't match other components disabled states.
>    image
>    vs
>    image

About the second point, here is a fix:
https://user-images.githubusercontent.com/15834481/143037276-9fe21288-49c7-45a7-b668-a1fbc2c99dc2.mov

If the user does not select a time and clicks outside the calendar, then it closes and the date is set to the selected date and the time to 00:00. How do you find this solution?

**Comment by sanmiawi** [comment]
> I found a bug while playing around with the date picker. See the video.
> 
>  Screen.Recording.2021-11-23.at.11.36.14.mov 
> Screenshot 2021-11-23 at 11 34 52
✅

**Comment by sanmiawi** [comment]
> I just noticed this strange behaviour. In my opinion, the regular expression works well and shows the "Invalid date format" error but the calendar underneath should not move to the next month. It's not any big bug, but I think we should keep a note in order to improve the UX during the enhancement phase.
> 
>  Screen.Recording.2021-11-23.at.11.10.17.mov

We can fix this by enabling `strictParsing` in the react-datepicker library but then we will not have the effect of the date being selected as a user is typing.

**Review by nicholas-os — APPROVED** [review]

**Review by Sakai-san — COMMENTED** [review]
I played around with the date picker example no 2. I did type in an `invalid` date, the pick a `time` in the right hand side column. Then I wanted to edit the time again. But you can't do that.

It s like the `input` field is `disabled`.

https://user-images.githubusercontent.com/6510794/143393957-d6a33c2a-c92e-4004-94c4-0350ee0c5deb.mov

**Review by Sakai-san — APPROVED** [review]

File: pyrene/src/components/DateTimePicker/DateTimePicker.tsx [hunk 19]
```diff
@@ -0,0 +1,303 @@
...
+   */
+  calendarOpened?: boolean,
+  /**
+   * Close calendar on date select
+   */
+  closeOnSelect?: boolean,
+  /**
+   * Date format used by component
+   */
+  dateFormat?: string,
```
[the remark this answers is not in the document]
sanmiawi: ✅

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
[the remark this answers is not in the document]
sanmiawi: ✅

pyrene/src/components/DateTimePicker/DateTimePicker.tsx
sanmiawi: 💡

PR #227, opened 2021-11-18
Tooltip adjust props type [improvement]

Sakai-san:


pyrene/src/components/Tooltip/Tooltip.tsx
```diff
@@ -23,7 +23,7 @@ export interface TooltipProps {
   /**
    * Sets the label displayed inside of the tooltip.
    */
-  label: React.ReactNode;
+  label: React.ReactNode | string;
```
cyrilwanner, line 26: Doesn't `ReactNode` already include `string`?

Sakai-san: @cyrilwanner you are right. Thanks !
Screenshot 2021-11-18 at 13 20 30
!

PR #228, opened 2021-11-19
typing duplicate [improvement]

Sakai-san:


cyrilwanner (approved):
That makes sense, thanks 👍

PR #229, opened 2021-11-22
Feature: Badge shows pointer cursor when clickable

maximelovino:
When `Badge` `onClick` function is defined, it shows a `pointer` cursor when hovered, otherwise it shows the default cursor


Sakai-san (approved):
LGTM. Thanks

cyrilwanner (approved):

PR #234, opened 2021-11-25
added 'disabled' state to table actions

sosostris:
Sometimes some of the table actions need to be disabled conditionally. Existing uses of this component has been giving the 'active' a 'disabled' value, and this was OK before the TS conversion as 'disabled' didn't match the predefined 'single', 'multi'  and 'always' and will return a 'false', now with TS it's stricter and 'disabled' gets warning. So we might as well just add this as one of the allowed strings.


Sakai-san (approved):
Thanks !

PR #235, opened 2021-11-25
checkbox popover click [bug]

Sakai-san:
# Done in this PR

Fix the `example`, such that a click `revert` back to the initial state.

___

https://user-images.githubusercontent.com/6510794/143459298-39a98f74-db2a-4510-acd2-11005e2237ae.mov


sosostris (approved):
Nice !

PR #236, opened 2021-11-29
fixed TreeTableCell display issue due to misplacement of '(' and ')'

sosostris:
The misplacement of the `()` led to that the space holder was always there


Sakai-san (approved):
Thank you so much @sosostris !

PR #238, opened 2021-11-30
FIX: Select title line height

maximelovino:
We need to specify line height as 12px for select title (as we do for
other form components). I think the `height` that was set here was a
typo and it was meant to be `line-height` which is why I replaced that
property.

This is important because otherwise it will follow line height
of whatever parent element so it will certainly look weird most of the
time unlike the other form components that specify it


Sakai-san (approved):

dparret (approved):
Good catch. It effectively should be 12px line-height. 
image

sanmiawi (approved):

PR #249, opened 2021-12-29
Migration of components from Kitchensink to Storybook

emandilara:
This PR contains the migration of the following components to `Storybook`:
- Banner
- Badge
- Card
- Key Value Table
- Simple Table

The existing stories are classified in categories, so that they are aligned with the `Kitchensink` examples hierarchy.

Screen capture:

https://user-images.githubusercontent.com/15834481/151551979-9069d0bf-3afa-411c-bf7c-c3e4451dcd33.mov


package-lock.json
```diff
@@ -910,7 +910,7 @@
       }
     },
     "@babel/plugin-transform-object-assign": {
-      "version": "file:pyrene/node_modules/@babel/plugin-transform-object-assign",
+      "version": "file:https:/registry.npmjs.org/@babel/plugin-transform-object-assign/-/plugin-transform-object-assign-7.14.5.tgz",
```
Sakai-san, line 913: @emandilara Hello. Thanks a lot for your work ! I think the changes that you have in the` package.json` are due to the following:

1. Fist you run `npm install` in the pyrene subproject.
2. Second you run `npm install` in the root directory.

emandilara: Hi @Sakai-san, thank you for your suggestion!

sanmiawi (approved):

cyrilwanner (commented):
I think we should not commit `pyrene/src/pyrene.min.css`, right? Or why was this added in this PR?

emandilara:
> I think we should not commit `pyrene/src/pyrene.min.css`, right? Or why was this added in this PR?

Good catch, thank you! I removed the lib.

cyrilwanner (approved):
Thank you, it looks great! 😃 And thank you for continuing the migration to Storybook 🙂

PR #252, opened 2021-12-30
[Fix] Correct lib import 

emandilara:
Correct component display names


cyrilwanner (approved):

cyrilwanner:
I think you also have to rename the file, in the [repository](https://github.com/open-ch/pyrene/tree/main/pyrene/src/components/Button), it is currently committed with a lowercase name.

emandilara:
> repository

Indeed - all the style imports will be updated in another PR, so I will actually discard this PR here.
In my local repo, having pull the latest changes from main, the file was with an uppercase first letter. I guess my branches got mixed up.

Thank you anyway for the review!

emandilara:
Apparently this was not taken into account because by default git ignores the change in capitalization of the filenames. I am taking this into account for the next PR that will update the filenames.

PR #256, opened 2022-01-07
Clean up unused styles

emandilara:
Cleaning up some CSS classes that do not have an implementation or are unused.

### DIFF
--- a/kitchensink/src/app/common/Search/ResultsPage.jsx
+++ b/kitchensink/src/app/common/Search/ResultsPage.jsx
@@ -37,7 +37,7 @@ export default class ResultsPage extends React.Component {
 
   renderSearchResults = () => (
     <>
-      <div className={styles.header}>
+      <div>
         <div className={styles.title}>{`${this.state.matches.length} result${this.state.matches.length > 1 ? 's' : ''} matching \u00AB${this.props.match.params.searchInput}\u00BB`}</div>
       </div>
       <div>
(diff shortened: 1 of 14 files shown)


kitchensink/src/app/static/foundations/IconsPage.jsx
```diff
@@ -32,7 +32,7 @@ const IconsPage = () => (
           Two-tone and multi-color icons are not available in the icon font. These icons are available as SVG assets.
         </DescriptionBox>
 
-        <div className="iconGrid">
```
cyrilwanner, line 35: Are you sure this is not used anymore?
On https://open-ch.github.io/pyrene/icons it looks like this class still adds some styles (looks like they are added globally here: https://github.com/open-ch/pyrene/blob/3ba1d1e37ebf2a7d6e249d38f632961246c716ba/kitchensink/src/css/common.css#L121) and the page looks kind of broken when I remove the class.

cyrilwanner (changes requested):
I think there is one issue with the icons page, but otherwise, it looks good to me.
Thanks for cleaning them up!

kitchensink/src/app/common/Search/ResultsPage.jsx
```diff
@@ -40,7 +40,7 @@ export default class ResultsPage extends React.Component {
       <div className={styles.header}>
```
sanmiawi, line 40: I guess this style can also be removed

kitchensink/src/app/static/foundations/IconsPage.jsx
emandilara: Indeed, thank you for pointing that out!

cyrilwanner (approved):

PR #257, opened 2022-01-24
Fix width button

sanmiawi:
Allow setting of button width.

Screenshot 2022-01-24 at 13 00 51
Screenshot 2022-01-24 at 13 01 58

### DIFF
--- a/pyrene/src/components/Button/Button.tsx
+++ b/pyrene/src/components/Button/Button.tsx
@@ -47,7 +47,7 @@ export interface ButtonProps {
    */
   type?: Type,
   /**
-   * Set button width
+   * Sets button width
    */
   width?: number,
 }
--- a/pyrene/src/components/Button/Button.tsx
+++ b/pyrene/src/components/Button/Button.tsx
@@ -46,6 +46,10 @@ export interface ButtonProps {
    * Sets the overall style.
    */
   type?: Type,
+  /**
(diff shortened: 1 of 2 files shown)


emandilara (approved):

pyrene/src/components/Button/Button.tsx
```diff
@@ -76,6 +81,7 @@ const Button: React.FC<ButtonProps> = ({
       }
       onClick={onClick}
       disabled={disabled}
+      style={{ width: width }}
```
maximelovino, line 84: this doesn't have a unit then, it's an error no? shouldn't it be 
```
style={{ width: `${width}px`}}
```
instead?

maximelovino (changes requested):

pyrene/src/components/Button/Button.tsx
maximelovino: Also, in this case you need to a check if `width` is undefined before as well

sanmiawi: If it is a `number` as specified by the type, React automatically adds the px, and if width is `undefined` then width is is not set.

maximelovino: Ok, that works for me

maximelovino (approved):

PR #258, opened 2022-01-25
Fix Select "create new tag". label

sanmiawi:
As discussed, extended the scope to include single select as both multi and single select implement the creatable prop which allows for the addition of new tags.

Screenshot 2022-01-26 at 07 52 23

### DIFF
--- a/pyrene/src/components/MultiSelect/MultiSelect.tsx
+++ b/pyrene/src/components/MultiSelect/MultiSelect.tsx
@@ -16,6 +16,10 @@ import { Option } from './types';
 
 export interface MultiSelectProps {
   /**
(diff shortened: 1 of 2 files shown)


maximelovino (approved):
LGTM

emandilara:
Nice @sanmiawi !
The prop name could stay as is, or could be changed into `newTagLabel` or `createTagLabel` for example.

The description could be something like this:
```
Helper label with information about the creation of a new tag, setting the text before its name.
```

emandilara (approved):

dparret:
What about `tagLabel` or `inputLabel` to keep some consistency with `helperLabel`?
Description is fine with me :)

dparret (approved):

PR #260, opened 2022-01-26
Don't show "no match found"

sanmiawi:
Select component should not show "No match found" if Select component has no default options and is `creatable`.


meluru:
Looks great :)

emandilara (approved):

PR #264, opened 2022-01-31
Fix fixed height

sanmiawi:
Issue: height of component is not affected if a rows prop of less than 3 is passed.

Fix

Screenshot 2022-01-31 at 09 51 55


emandilara (approved):

pyrene/src/components/TextArea/textArea.css
```diff
@@ -74,7 +73,7 @@
 
   font-family: FiraGO, Helvetica, sans-serif;
   font-size: 12px;
-  line-height: 12px;
+  line-height: 14px;
```
dparret, line 76: It makes it to 32px with 1 row only (that's good), but it doesn't feel right, because when we add up multiple rows, the size of the whole is not a multiple of 8 anymore.
For me, we should have it to 16px, and padding 7 (+1 for the border).

dparret (changes requested):
Negative numbers (and 0) should either not be valid, or behave like 1 imo.

pyrene/src/components/TextArea/textArea.css
sanmiawi: ✅

sanmiawi:
> Negative numbers (and 0) should either not be valid, or behave like 1 imo.
✅

pyrene/src/components/TextArea/TextArea.tsx
```diff
@@ -112,7 +112,7 @@ const TextArea: React.FC<TextAreaProps> = ({
         className={clsx(styles.textArea, { [styles.resizeable]: resizeable }, { [styles.filled]: value })}
         name={name}
         placeholder={placeholder}
-        rows={rows}
+        rows={rows < 1 ? 1 : rows}
```
maximelovino, line 115: I would suggest using `rows={Math.min(1, rows)}`

sanmiawi: > Negative numbers (and 0) should either not be valid, or behave like 1 imo.

The idea as stated by @dparret is to equate to 1 row if the number of `rows` passed is less than or equal to 0

> Math.min(1, -5)

This will return -5 instead of 1 as expected

dparret (commented):
Thanks for the update, good to me.

dparret (approved):

pyrene/src/components/TextArea/TextArea.tsx
sanmiawi: ✅

PR #266, opened 2022-01-31
[textarea] Add adaptToContent and maxRows properties

meluru:
Another approach, much better imho


dparret (approved):
Works well locally.

pyrene/src/components/TextArea/TextArea.tsx
```diff
@@ -89,10 +101,39 @@ const TextArea: React.FC<TextAreaProps> = ({
...
   const characterCount = maxLength - (value !== null ? value.length : 0);
   const characterLimitReached = characterCount < 0;
+  const textAreaRef = useRef<HTMLTextAreaElement>(null);
+  const [currentRows, setCurrentRows] = useState(minRows);
+  const [text, setText] = useState(value);
+
+  useEffect(() => {
+    if (adaptToContent && textAreaRef.current) {
+      // eslint-disable-next-line no-bitwise
+      const rowsFromHeight = ~~(textAreaRef.current.scrollHeight / textAreaLineHeight);
```
sanmiawi, line 119: I think `Math.floor()` is easier to read and `~~` starts to fail when dividing very large numbers.

meluru: That makes sense, thanks :) Will fix it

pyrene/src/components/TextArea/TextArea.tsx
```diff
@@ -89,10 +101,39 @@ const TextArea: React.FC<TextAreaProps> = ({
...
+  const textAreaLineHeight = 16;
   const characterCount = maxLength - (value !== null ? value.length : 0);
   const characterLimitReached = characterCount < 0;
+  const textAreaRef = useRef<HTMLTextAreaElement>(null);
+  const [currentRows, setCurrentRows] = useState(minRows);
+  const [text, setText] = useState(value);
+
+  useEffect(() => {
+    if (adaptToContent && textAreaRef.current) {
+      // eslint-disable-next-line no-bitwise
```
sanmiawi, line 118: I guess this can also be removed this since we are no longer doing the bitwise operation

PR #268, opened 2022-02-07
Multiselect input delimiter

sanmiawi:
Keep Default delimiters still as it is now.

Implement a prop such as `customDelimiters: ['\\s', ',', ';', '?']`

We want to make sure a user can type in 

- something, somethingelse and have 2 entries
- something somethingelse and have 2 entries


maximelovino (approved):
LGTM, also tested it locally. Good idea to emulate the enter key to trigger new words in react-select 😁

emandilara (approved):

PR #270, opened 2022-02-08
Select should keep content when clicking outside

sanmiawi:
Screenshot 2022-02-07 at 18 01 13


dparret (approved):
Works well locally for me. If the tag isn't created, is it considered as an input value? I don't think it should. WDYT?

meluru (approved):

